PR #16, opened 2021-02-19
Add namespace metadata to resources

abhilashjoseph:
Hello,

This pull request is to add namespace metadata to resources that this chart generates so that the resources go to the desired namespace instead of the namespace where the job / process is running.


abhilashjoseph:
Hello, Can someone please have a look into this PR, need these changes badly in our deployment, thanks in advance!

erkannt:
Haven't come across `Release.Namespace` before, so did some reading.

https://github.com/helm/helm/issues/5465#issuecomment-723134829

One of the Helm maintainers answers to that comment. According to him this PR would go against best practice of _not_ specifying a namespace via the chart.

As of now I'd suggest not merging this PR as I don't understand how this might affect existing workflows that use this chart. What happens if they use helm in a way that leads to an empty `Release.Namespace`? From the [helm docs](https://helm.sh/docs/chart_template_guide/builtin_objects/) it is unclear to me how this value gets set, not-set, overridden.

@abhilashjoseph Could you maybe shed some light here? I'd be keen to understand why your current workflow would require this change.

tlawrie:
Hi @erkannt yes I agree. From my experience crafting Helm charts its an unwritten rule (although may now be written) that you don't add namespace. Thats taken care of by the helm commands in determining which namespace you want to install it into.

abhilashjoseph:
@erkannt, This is the reason https://github.com/helm/helm/issues/5465#issuecomment-723536030 why we need this capability. Also as per the link shared earlier, it talks about hard coding namespaces or getting the value for namespace from values as "not recommended practice".
It is a general practice to send manifests to specific namespaces and default to the release namespace instead of them going into the default namespace.

abhilashjoseph:
The specific reason why we need this is, because oauth2proxy is a dependency (helm) to another Chart which configures ingress gateways using istio and has multiple related components that go into specific namespaces. Overriding /forcing the kubectl apply with a namespace causes errors in running jobs (since the jobs aren't being created in proper namespaces) and ends up with failed deployments. This way of handling the namespace or defaulting to the release namespace is done by almost all major helm charts. It would be really beneficial to get this out as quickly as possible.

desaintmartin:
This looks like a hack, but I've seen this before.
What are the resources going into the default namespace? Maybe we can properly solve the problem by changing a few other lines.

SeWieland:
@abhilashjoseph have you thought about not making oauth2proxy a direct dependency?
I only use chart dependecies if the the resources really belong very closely together.

You could as well use something like [helmfile](https://github.com/roboll/helmfile) and specify the namespaces + the correct order where / when helm releases need to be placed exactly (at least that is how am I handling these kinds of issues)..

abhilashjoseph:
> @abhilashjoseph have you thought about not making oauth2proxy a direct dependency?
> I only use chart dependecies if the the resources really belong very closely together.
> 
> You could as well use something like [helmfile](https://github.com/roboll/helmfile) and specify the namespaces + the correct order where / when helm releases need to be placed exactly (at least that is how am I handling these kinds of issues)..

@SeWieland, thank you for your comments, In my case the outh2proxy is actually a very direct dependency to the extra set of charts we deploy for Istio deployment. And yes splitting the chart to it's own chart is a possibility.

pierluigilenoci:
@abhilashjoseph is this PR still useful to you?

abhilashjoseph:
> @abhilashjoseph is this PR still useful to you?

@pierluigilenoci PR is still useful, but doesn't look like this is going to make into main, so closing it!

PR #17, opened 2021-02-22
fix: get chart-testing to lint the charts

erkannt:
The config doesn't like the `./` in specifying the charts directory.

Looking at the current actions you can see that the linter is skipping all the directories.

I could repro this locally and got it fixed with the submitted change.

@JoelSpeed

### DIFF
--- a/ct.yaml
+++ b/ct.yaml
@@ -1,7 +1,7 @@
 # See https://github.com/helm/chart-testing#configuration
 remote: origin
 chart-dirs:
-  - ./helm
+  - helm
 target-branch: main
 helm-extra-args: --timeout 600s
 


JoelSpeed (approved):
Would it be worth pushing a dummy change, checking the linter detects the change and then removing it?

erkannt:
I've merged this PR in my fork of the repo and then created a PR to try get the linter to fail.

- comparison of [my fork](https://github.com/oauth2-proxy/manifests/compare/main...erkannt:main) and main
- the [PR failing linting](https://github.com/erkannt/manifests/pull/2)

The first run failed due to the chart having been modified but the version not being bumped. Once I bumped the version it picked up on the key I had deleted.

PR #20, opened 2021-03-09
add bitnami to chart-repos for ct

erkannt:
required by #7

#19 is also required to get the test pipeline green in general


JoelSpeed (approved):
Seems good

PR #28, opened 2021-05-12
[PR] Fixed OAUTH2_PROXY_REDIS_PASSWORD env variable

pierluigilenoci:
The chart, even in the presence of `existingSecret`, still used the name of the default secret.


desaintmartin (approved):
Thanks!

PR #37, opened 2021-06-17
Added ingress for apiVersion: networking.k8s.io/v1

### LINKED ISSUE #35: Ingress apiVersions will be obsolete shortly
Hi

Ingress apiVersions: [extensions/v1beta1 & networking.k8s.io/v1beta1](https://github.com/oauth2-proxy/manifests/blob/24ced0a77e4fd38e6ba8080f2eb85b89a4980528/helm/oauth2-proxy/templates/ingress.yaml#L7) are [deprecated](https://kubernetes.io/docs/reference/using-api/deprecation-guide/#ingress-v122) and will fail to deploy in the next version of Kubernetes (v1.22 - Aug 2021).  Should be replaced by networking.k8s.io/v1.

Thanks.

SeWieland:
Adds ingress changes needed for "networking.k8s.io/v1".

Closes #35 

I just made the basic changes and have not yet tested it!

### DIFF
--- a/helm/oauth2-proxy/Chart.yaml
+++ b/helm/oauth2-proxy/Chart.yaml
@@ -1,5 +1,5 @@
 name: oauth2-proxy
-version: 3.3.3
+version: 4.0.0
 apiVersion: v2
 appVersion: 7.1.3
 home: https://oauth2-proxy.github.io/oauth2-proxy/
--- a/helm/oauth2-proxy/Chart.yaml
+++ b/helm/oauth2-proxy/Chart.yaml
@@ -1,5 +1,5 @@
 name: oauth2-proxy
-version: 3.3.2
+version: 3.3.3
 apiVersion: v2
 appVersion: 7.1.3
 home: https://oauth2-proxy.github.io/oauth2-proxy/
--- a/helm/oauth2-proxy/README.md
+++ b/helm/oauth2-proxy/README.md
@@ -46,11 +46,19 @@ This version upgrades oauth2-proxy to v4.0.0. Please see the [changelog](https:/
 
 Version 2.0.0 of this chart introduces support for Kubernetes v1.16.x by way of addressing the deprecation of the Deployment object apiVersion `apps/v1beta2`.  See [the v1.16 API deprecations page](https://kubernetes.io/blog/2019/07/18/api-deprecations-in-1-16/) for more information.
 
-Due to [this issue](https://github.com/helm/helm/issues/6583) there may be errors performing a `helm upgrade`of this chart from versions earlier than 2.0.0.
+Due to [this issue](https://github.com/helm/helm/issues/6583) there may be errors performing a `helm upgrade` of this chart from versions earlier than 2.0.0.
 
 ### To 3.0.0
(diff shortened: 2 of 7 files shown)


pierluigilenoci:
@SeWieland could you please test it?

SeWieland:
@pierluigilenoci Done.

I created the new release to test under the fork https://sewieland.github.io/manifests and it worked seamlessly under my setup with the new ingress API (K8s Rev v1.20.5) 👍 ([chart release](https://github.com/SeWieland/manifests/releases/download/oauth2-proxy-3.3.3/oauth2-proxy-3.3.3.tgz) if someone wants to have a look at it for backwards compatibility)

SeWieland:
There is still one more issue:

If for any reason someone declared `service.port` in his values with a named port (string) and already has the new api capability, the chart will break as the new API distinguishes between `port.number` and `port.name`.
However, renaming the value to `service.portNumber` will break _all_ setups that specify a custom port.

Should this be mentioned somewhere?

holmesb:
Default is number (80) not name.  I think this is an edge-case we should mention in the commit message, readme anywhere else possible during the release and then proceed.  Without this merged, this chart will break for *everyone* on the latest kubernetes from the start of August.

SeWieland:
True that!
The default is already a port 80 - so I really doubt if there is anyone at all experiencing this issue after the update.

pierluigilenoci (dismissed):
LGTM

pierluigilenoci:
@desaintmartin this is my first PR as a reviewer, could you please take a look?

helm/oauth2-proxy/Chart.yaml
```diff
@@ -1,5 +1,5 @@
 name: oauth2-proxy
-version: 3.3.2
+version: 3.3.3
```
desaintmartin, line 2: I think this is a breaking change, we should advertise it (major version bump + requiring kubernetes 1.xx in chart.yaml + readme + explain how to upgrade in the README).

People with existing Releases with `ingress.extraPaths` values will get impacted and need to adapt their values.

desaintmartin (changes requested):

helm/oauth2-proxy/Chart.yaml
pierluigilenoci: I totally agree

morganchristiansson: Can check and fail if ingress.extraPaths has old format..
See example from jenkins chart https://github.com/jenkinsci/helm-charts/blob/main/charts/jenkins/templates/deprecation.yaml

SeWieland: I like the deprecation.yaml approach :+1:
I will propose one in this MR and do a version bump to v4.0.0

SeWieland:
I added the deprecation check and can confirm that it works as expected.
Please review again

pierluigilenoci (approved):
LGTM but I'll wait for @desaintmartin to merge 😘

desaintmartin (approved):
Thanks!

PR #39, opened 2021-06-29
Uses template function to generate service account name

caiofralmeida:
I would like to override the service account name instead use the default full name.

Signed-off-by: Caio Almeida <caio.almeida@gympass.com>

PR #40, opened 2021-06-30
Make compatible with K8s 1.18

curamos:
# Kubernetes 1.18:
$ kubectl api-versions | grep networking
networking.k8s.io/v1
networking.k8s.io/v1beta1
$ kubectl explain --api-version=networking.k8s.io/v1 Ingress
error: Couldn't find resource for "networking.k8s.io/v1, Kind=Ingress"

### DIFF
--- a/helm/oauth2-proxy/templates/ingress.yaml
+++ b/helm/oauth2-proxy/templates/ingress.yaml
@@ -5,10 +5,10 @@
 {{- $ingressPathType := .Values.ingress.pathType -}}
 {{- $extraPaths := .Values.ingress.extraPaths -}}
 {{- $apiV1 := false -}}
-{{- if .Capabilities.APIVersions.Has "networking.k8s.io/v1" -}}
+{{- if .Capabilities.APIVersions.Has "networking.k8s.io/v1/Ingress" -}}
 apiVersion: networking.k8s.io/v1
 {{- $apiV1 = true -}}
-{{- else if .Capabilities.APIVersions.Has "networking.k8s.io/v1beta1" -}}
+{{- else if .Capabilities.APIVersions.Has "networking.k8s.io/v1beta1/Ingress" -}}
 apiVersion: networking.k8s.io/v1beta1
 {{- else -}}
 apiVersion: extensions/v1beta1
--- a/helm/oauth2-proxy/Chart.yaml
+++ b/helm/oauth2-proxy/Chart.yaml
@@ -1,5 +1,5 @@
 name: oauth2-proxy
-version: 4.0.2
+version: 4.0.3
 apiVersion: v2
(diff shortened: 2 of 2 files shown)


caiofralmeida:
@curamos When I did a test putting the resource name on the conditions, isn't worked for me. Because of that, I opened a PR that parameterizes the apiVersion via values https://github.com/oauth2-proxy/manifests/pull/41.

morganchristiansson:
use chart version 3.3.2 as workaround

caiofralmeida:
@morganchristiansson thanks, but I need to change the service account name, I opened this PR https://github.com/oauth2-proxy/manifests/pull/39 to fix it. I think this but will not fix it in chart version 3.3.2, right?

pierluigilenoci:
I find this fix better than specifying the API version directly in the values file (ref: https://github.com/oauth2-proxy/manifests/pull/41).

Could you also bump the chart and squash all in a single commit? 

@desaintmartin what do you think about it?

helm/oauth2-proxy/templates/ingress.yaml
```diff
@@ -5,7 +5,7 @@
 {{- $ingressPathType := .Values.ingress.pathType -}}
 {{- $extraPaths := .Values.ingress.extraPaths -}}
 {{- $apiV1 := false -}}
-{{- if .Capabilities.APIVersions.Has "networking.k8s.io/v1" -}}
+{{- if .Capabilities.APIVersions.Has "networking.k8s.io/v1/Ingress" -}}
 apiVersion: networking.k8s.io/v1
 {{- $apiV1 = true -}}
 {{- else if .Capabilities.APIVersions.Has "networking.k8s.io/v1beta1" -}}
```
pierluigilenoci, line 11: Should `/Ingress` be specified also here?

pierluigilenoci:
> @morganchristiansson thanks, but I need to change the service account name, I opened this PR #39 to fix it. I think this but will not fix it in chart version 3.3.2, right?

Of course not. 😄

helm/oauth2-proxy/templates/ingress.yaml
pierluigilenoci: I'd prefer to keep this PR as complete as possible. 😜 
For me, it makes sense to check inside the Ingress manifest if ".Capabilities.APIVersions.Has" for the Ingress specific API in the whole file, not only in one point.

pierluigilenoci:
The chart needs another version bump.

helm/oauth2-proxy/templates/ingress.yaml
curamos: Changed it.

pierluigilenoci (approved):

PR #41, opened 2021-06-30
Templating ingress api version to support Kubernetes 1.18

caiofralmeida:
We are not able to use the helm chart because our Kubernetes cluster is running in version 1.18. 

Explanation

I believe Kubernetes 1.18 already supports the API version `networking.k8s.io/v1`, but only for the `NetworkPolicy` resource (https://kubernetes.io/docs/reference/using-api/deprecation-guide/ #networkpolicy-v116) and in the case of the Ingress resource it will only be available in version 1.19 (https://kubernetes.io/docs/reference/using-api/deprecation-guide/#ingress-v122).

Showing the API versions available:
```
$ kubectl api-versions | grep networking.k8s.io
networking.k8s.io/v1
networking.k8s.io/v1beta1
```

Because our Kubernetes cluster has the `networking.k8s.io/v1` API, it ends up falling into this condition, but we don't have the `networking.k8s.io/v1/Ingress` feature.
```
{{- if .Capabilities.APIVersions.Has "networking.k8s.io/v1" -}}
apiVersion: networking.k8s.io/v1
...
```
Showing  the API resources for ingresses:
```
$ kubectl api-resources | grep ingresses
ingresses                                             ing                extensions/v1beta1                  true         Ingress
ingresses                                             ing                networking.k8s.io/v1beta1           true         Ingress
```
The idea of this PR is to provide as an option the override of the Ingress apiVersion via `values.yaml`.

### DIFF
--- a/helm/oauth2-proxy/Chart.yaml
+++ b/helm/oauth2-proxy/Chart.yaml
@@ -1,5 +1,5 @@
 name: oauth2-proxy
-version: 4.0.2
+version: 4.0.3
 apiVersion: v2
 appVersion: 7.1.3
 home: https://oauth2-proxy.github.io/oauth2-proxy/
--- a/helm/oauth2-proxy/Chart.yaml
+++ b/helm/oauth2-proxy/Chart.yaml
@@ -1,5 +1,5 @@
 name: oauth2-proxy
-version: 4.0.3
+version: 4.0.4
 apiVersion: v2
 appVersion: 7.1.3
 home: https://oauth2-proxy.github.io/oauth2-proxy/
--- a/helm/oauth2-proxy/README.md
+++ b/helm/oauth2-proxy/README.md
@@ -97,7 +97,6 @@ Parameter | Description | Default
 `ingress.enabled` | Enable Ingress | `false`
 `ingress.path` | Ingress accepted path | `/`
 `ingress.pathType` | Ingress [path type](https://kubernetes.io/docs/concepts/services-networking/ingress/#path-types) | `ImplementationSpecific`
-`ingress.apiVersion` | Ingress api version | `nil`
 `ingress.extraPaths` | Ingress extra paths to prepend to every host configuration. Useful when configuring [custom actions with AWS ALB Ingress Controller](https://kubernetes-sigs.github.io/aws-alb-ingress-controller/guide/ingress/annotation/#actions). | `[]`
 `ingress.annotations` | Ingress annotations | `nil`
 `ingress.hosts` | Ingress accepted hostnames | `nil`
--- a/helm/oauth2-proxy/README.md
+++ b/helm/oauth2-proxy/README.md
@@ -96,6 +96,8 @@ Parameter | Description | Default
 `imagePullSecrets` | Specify image pull secrets | `nil` (does not add image pull secrets to deployed pods)
 `ingress.enabled` | Enable Ingress | `false`
 `ingress.path` | Ingress accepted path | `/`
+`ingress.pathType` | Ingress [path type](https://kubernetes.io/docs/concepts/services-networking/ingress/#path-types) | `ImplementationSpecific`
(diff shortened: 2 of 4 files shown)


pierluigilenoci:
I'm not a big fan of this kind of thing. It would be better to use `Capabilities.APIVersions.Has`

caiofralmeida:
@pierluigilenoci It makes sense about `Capabilities.APIVersions.Has` (https://helm.sh/docs/chart_template_guide/builtin_objects/), I went for this approach as testing the resource condition locally didn't work, but maybe I did something wrong.

pierluigilenoci:
@caiofralmeida I understand you had some problems but we need to find a way to make it work. 😜 

What version of helm do you use?

pierluigilenoci:
@caiofralmeida you need to fix the chart bump and resolve the conflict. 😉

caiofralmeida:
> @caiofralmeida I understand you had some problems but we need to find a way to make it work. 😜
> 
> What version of helm do you use?

Hi @pierluigilenoci, I did tests locally using helm 3.3 and 3.6 within the command `helm template`
 I've been thinking if it's work running `helm install --dry-run` command, because that contacts the server.

Thank you for all support 🙂

caiofralmeida:
@pierluigilenoci confirming what I expected. When I run only the command `helm template` isn't work, but using `helm install --dry-run` works perfectly, sorry for the confusion.

More details:

Given the current ingress template is:
```yaml
{{- if .Values.ingress.apiVersion -}}
apiVersion: {{ .Values.ingress.apiVersion }}
{{- else if .Capabilities.APIVersions.Has "networking.k8s.io/v1/Ingress" -}}
apiVersion: networking.k8s.io/v1
{{- $apiV1 = true -}}
{{- else if .Capabilities.APIVersions.Has "networking.k8s.io/v1beta1/Ingress" -}}
apiVersion: networking.k8s.io/v1beta1
{{- else -}}
apiVersion: extensions/v1beta1
{{- end }}
kind: Ingress
...
```
Using helm 2 / helm 3.3 / helm 3.6 (helm template --values values-staging.yaml -x templates/ingress.yaml --name foo .)
```yaml
# Source: oauth2-proxy/templates/ingress.yaml
apiVersion: extensions/v1beta1
kind: Ingress
...
```
Using helm 3.3 (helm3.3 install foo --values values-staging.yaml  . --dry-run)
```
# Source: oauth2-proxy/templates/ingress.yaml
apiVersion: networking.k8s.io/v1beta1
kind: Ingress
```
Therefore, I think that this PR (#41) that opens no longer makes sense because as you commented, this one (#40) already solves, what do you think?

pierluigilenoci:
@caiofralmeida it doesn't matter if we merge #40 or #41, the first PR who respects all the requirements wins. 😜 

On the other side I don't like that much this https://github.com/oauth2-proxy/manifests/pull/40#pullrequestreview-696173116 so feel free to change your PR to comply 😉

caiofralmeida:
> @caiofralmeida it doesn't matter if we merge #40 or #41, the first PR who respects all the requirements wins. 
> 
> On the other side I don't like that much this [#40 (review)](https://github.com/oauth2-proxy/manifests/pull/40#pullrequestreview-696173116) so feel free to change your PR to comply

@pierluigilenoci could you check, please if this PR is complying?

I took the opportunity to update Ingress [documentation](https://github.com/oauth2-proxy/manifests/pull/41/files#diff-0a2e8bbdc6a428423422f800e4678c6cef89d48c6016cfbd88bcdc6bb6895570R99), but I don't know if I should continue in this PR.

pierluigilenoci (approved):

pierluigilenoci:
@caiofralmeida I took the liberty of changing the PR a bit because this fix to the README file was needed anyway.

pierluigilenoci (approved):

PR #49, opened 2021-07-22
Update ingress.yaml: support empty list for .Values.ingress.hosts

SleepyBag:
For now, if the hosts field is not set, there will be an error while running `helm install`.

However, the host field is not forced to set in Kubernetes. No host field just means any host is accepted.

I update ingress.yaml so it won't throw error for empty hosts list then.

### DIFF
--- a/helm/oauth2-proxy/templates/ingress.yaml
+++ b/helm/oauth2-proxy/templates/ingress.yaml
@@ -27,9 +27,31 @@ metadata:
 {{- end }}
 spec:
   rules:
+    {{- if .Values.ingress.hosts }}
     {{- range $host := .Values.ingress.hosts }}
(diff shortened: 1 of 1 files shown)


pierluigilenoci:
@SleepyBag this PR needs to be fixed, several things are missing. One thing above all, the chart version bump.

@desaintmartin what do you think?

desaintmartin:
This is a very complicated way to state that host is optional, isn't it? Moreover, I am not sure we want to allow a wildcard ingress (for reference, it seems the helm default template when creating a chart does not either).

pierluigilenoci:
@SleepyBag so?

pierluigilenoci:
@SleepyBag any update?

PR #52, opened 2021-08-06
Set Ingress network API version based on k8s version

### LINKED ISSUE #51: Ingress has wrong API version with k8s 1.19
Applying chart version 4.1.0 on k8s 1.19 results in the wrong API version for the Ingress object:
```
extensions/v1beta1
```
instead of
```
networking.k8s.io/v1
```

With chart version 4.0.2 it works as expected.

This is the commit which changed the behavior: https://github.com/oauth2-proxy/manifests/commit/9ef089b45922650840cdf4c44842c7a2e9691151#diff-d2ec12fa131797b240c1ae945cb0fb458d289bca1759b6839a219928860baad2

There was already some discussion on how to implement this: https://github.com/oauth2-proxy/manifests/pull/41

czunker:
Fixes #51

### DIFF
--- a/helm/oauth2-proxy/Chart.yaml
+++ b/helm/oauth2-proxy/Chart.yaml
@@ -1,17 +1,17 @@
 name: oauth2-proxy
-version: 4.2.0
+version: 4.2.1
 apiVersion: v2
 appVersion: 7.1.3
 home: https://oauth2-proxy.github.io/oauth2-proxy/
 description: A reverse proxy that provides authentication with Google, Github or other providers
 keywords:
-- kubernetes
-- oauth
-- oauth2
-- authentication
-- google
-- github
-- redis
+  - kubernetes
+  - oauth
+  - oauth2
+  - authentication
+  - google
+  - github
+  - redis
 dependencies:
   - name: redis
     version: ~10.6.0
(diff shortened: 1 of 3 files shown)


morganchristiansson:
commit message and PR title has misspelled verison

czunker:
> commit message and PR title has misspelled verison

Fixed

pierluigilenoci:
@caiofralmeida @curamos @SeWieland @desaintmartin what do you think of this PR?
I would like to find a shared solution to this problem.

mattfield:
Passing +1 to this PR from a user of the chart in production. Currently we're running a local patched copy to deal with this issue, but would love to get rid of that in favour of referencing a fixed version from upstream.

desaintmartin:
For example, our friends at Bitnami only compare version instead of features availability:
- https://github.com/bitnami/charts/blob/master/bitnami/common/templates/_capabilities.tpl#L65
- https://github.com/bitnami/charts/blob/master/bitnami/wordpress/templates/ingress.yaml

I suppose we could even drop the first `and` but it should anyway work like this!

czunker:
Fixed linter issue.

desaintmartin (approved):
Thanks!

desaintmartin:
@pierluigilenoci did you have something specific in mind when writing about generic solution?

pierluigilenoci:
@desaintmartin maybe you are getting confused with another issue.
I wrote "shared solution" meaning that we all had to approve the proposed solution. 
This is not the first PR that tries to solve this problem.
Instead of continuing to try, I thought it was appropriate to find a solution altogether. 😄
For example, I really like the way the guys at Bitnami solve the problem.

desaintmartin:
I suppose we could mimick (or even use their `common` Chart?) the way Bitnami solves the problem!
For now I propose we merge this PR, is that OK?

pierluigilenoci:
@desaintmartin ok to merge from my side.

PR #53, opened 2021-08-06
Correct table formatting for `cookie.cookieSecret` row

gchappell99:
The embedded pipes in the example `openssl` command for setting `cookie.cookieSecret` were causing the end of that command to overflow into the next column making the command incomplete/non-functional.

### DIFF
--- a/helm/oauth2-proxy/Chart.yaml
+++ b/helm/oauth2-proxy/Chart.yaml
@@ -1,5 +1,5 @@
 name: oauth2-proxy
-version: 4.1.0
+version: 4.1.1
 apiVersion: v2
 appVersion: 7.1.3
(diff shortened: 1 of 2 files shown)


pierluigilenoci:
@gchappell99 could you please also add a chart bump?

gchappell99:
> @gchappell99 could you please also add a chart bump?

I assume you mean bump the minor version in `Chart.yaml`; I didn't think it would matter for a doc change vs a functionality change but I'll do that right now

pierluigilenoci:
@gchappell99 it also sounds strange to me to bump for a change to the documentation but that's how it works. Dura lex, sed lex. 😉

pierluigilenoci (approved):

Issue #60, opened 2021-10-01
Unable to use ingress v1 on GKE


gmenuel:
On GKE, and probably other kubernetes versions, the semver returned by .Capabilities.KubeVersion.Version contains a prerelease string (eg the version could be : v1.20.9-gke.1001). In Helm doing a semverCompare ">= v1.19.0" "v1.20.9-gke.1001" will be false, because semverCompare considers prerelease to be invalid if the version constraint doesn't include a prerelease (see https://github.com/Masterminds/semver#checking-version-constraints ).

This prevents GKE user of the oauth2-proxy chart from using Ingress with API networking.k8s.io/v1. So fix is simple, the semver check in https://github.com/oauth2-proxy/manifests/blob/main/helm/oauth2-proxy/templates/ingress.yaml#L8 should be: `semverCompare ">= v1.19.0-0" .Capabilities.KubeVersion.Version`

pierluigilenoci:
Thank you @gmenuel, feel free to make a PR with this fix.

PR #61, opened 2021-10-01
Correct semverCompare constraints for ingress. Fixes: #60

gmenuel:
Fix the user of ingress api v1 on GKE, and probably other kubernetes versions. See: #60


desaintmartin:
Thanks! See the failing test logs (version bump needed).

gmenuel:
Yeah I figured that the chart version should probably be changed. It should be OK now.

PR #64, opened 2021-10-26
Service monitor labels

nlamirault:


pierluigilenoci (dismissed):

pierluigilenoci:

helm/oauth2-proxy/templates/servicemonitor.yaml
```diff
@@ -10,10 +10,7 @@ metadata:
 {{- end }}
   labels:
     app: {{ template "oauth2-proxy.name" . }}
```
desaintmartin, line 12: Why do you keep this one?

helm/oauth2-proxy/values.yaml
```diff
@@ -259,4 +259,5 @@ metrics:
     # Prometheus scrape timeout
     scrapeTimeout: 30s
     # Add custom labels to the ServiceMonitor resource
-    labels: {}
+    labels:
+      prometheus: default
```
desaintmartin, line 263: This should probably be empty by default, don't you think?

helm/oauth2-proxy/Chart.yaml
```diff
@@ -1,5 +1,5 @@
 name: oauth2-proxy
-version: 5.0.2
+version: 5.0.3
```
pierluigilenoci, line 2: This version is already used.

helm/oauth2-proxy/templates/servicemonitor.yaml
pierluigilenoci: The only reason I see is to keep the chart backward compatible with installations using this label.
At least we can add a note to the release because it is not a real breaking change.

helm/oauth2-proxy/values.yaml
pierluigilenoci: I asked for it as not to change the default behavior. That value came from here. 
https://github.com/oauth2-proxy/manifests/blob/main/helm/oauth2-proxy/values.yaml#L261
and here
https://github.com/oauth2-proxy/manifests/blob/main/helm/oauth2-proxy/templates/servicemonitor.yaml#L16

nlamirault:
Fix by : https://github.com/oauth2-proxy/manifests/pull/72

PR #48, opened 2021-07-21
Fix Google serviceAccountJson

skirsten:
The helm template for the `config.google.serviceAccountJson` was not working before.

How to use:
```yaml
config:
  google:
    adminEmail: "admin@my.com"
    serviceAccountJson: |
      {
        "type": "service_account",
        ...
      }
```

PR #65, opened 2021-10-27
fix google service-account.json secret generation

maxlaverse:
I spend some time wrapping my head around this since it has been in the repository for quite some time, and I came to the conclusion that it's broken. I suppose not many people use membership verification ?

As a result you would get a `Secret` without value, which when applying with Terraform ends up in:
```
Error: unable to build kubernetes objects from release manifest: error validating "": error validating data: unknown object type "nil" in Secret.data.service-account.json
```

### DIFF
--- a/helm/oauth2-proxy/Chart.yaml
+++ b/helm/oauth2-proxy/Chart.yaml
@@ -1,5 +1,5 @@
 name: oauth2-proxy
-version: 5.0.2
+version: 5.0.3
 apiVersion: v2
 appVersion: 7.2.0
 home: https://oauth2-proxy.github.io/oauth2-proxy/
--- a/helm/oauth2-proxy/Chart.yaml
+++ b/helm/oauth2-proxy/Chart.yaml
@@ -1,5 +1,5 @@
 name: oauth2-proxy
-version: 5.0.3
+version: 5.0.4
 apiVersion: v2
 appVersion: 7.2.0
 home: https://oauth2-proxy.github.io/oauth2-proxy/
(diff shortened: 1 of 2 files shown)


pierluigilenoci:
@maxlaverse this PR needs to be fixed, several things are missing. One thing above all, the chart version bump.

Besides the fact that I see it linked to #48, it might make sense to make an overall of the changes needed to authenticate via Google.

maxlaverse:
@pierluigilenoci I pushed a commit bumping the Chart version to 5.0.3.

PR  https://github.com/oauth2-proxy/manifests/pull/48 was also changing the `secretName` value but this is on `master` if I read it right. The only remaining difference is that this PR doesn't encode the `serviceAccountJSON` in base64 and I would keep it that way, since the documentation doesn't mention this encoding. 

If we'd like to encode the content of the JSON, I can push another commit and update the documentation.

pierluigilenoci:
@desaintmartin what you think about?

desaintmartin:
Indeed, this was completely wrong.

Actually, you'll need to encode in b64 (unless you use `stringData`), or Kubernetes will complain.

maxlaverse:
Completely forgot the Secret needed to be base64 encoded in any case. Fixed @pierluigilenoci

helm/oauth2-proxy/Chart.yaml
```diff
@@ -1,5 +1,5 @@
 name: oauth2-proxy
-version: 5.0.2
+version: 5.0.3
```
pierluigilenoci, line 2: Version needs to be bumped, 5.0.3 is already used.

pierluigilenoci (approved):

pierluigilenoci:
@desaintmartin I'm not an expert on that, can you take a look?

desaintmartin:
Thanks!

desaintmartin (approved):

PR #66, opened 2021-11-10
[oauth2-proxy/PDB] Add Kubernetes 1.21+ support & fix broken selector labels

julien-sugg:
Changeset:

feat: update PDB to support Kubernetes 1.21+ in a backward compatible fashion - Starting Kubernetes 1.21, the PDB feature went stable
fix: rework and fix broken PDB selector labels - It appears to have broken a long time ago and is related to 0ea4dc4a037a803191b03709896f08f689be6444 and cbd5275649b7aab0a1c01f71895e707460e5c607
chore: bump chart version (patch)

### DIFF
--- a/helm/oauth2-proxy/Chart.yaml
+++ b/helm/oauth2-proxy/Chart.yaml
@@ -1,5 +1,5 @@
 name: oauth2-proxy
-version: 5.0.2
+version: 5.0.3
 apiVersion: v2
 appVersion: 7.2.0
 home: https://oauth2-proxy.github.io/oauth2-proxy/
--- a/helm/oauth2-proxy/templates/_capabilities.tpl
+++ b/helm/oauth2-proxy/templates/_capabilities.tpl
@@ -8,3 +8,16 @@ Returns the appropriate apiVersion for podDisruptionBudget object.
 {{- print "policy/v1beta1" -}}
 {{- end -}}
 {{- end -}}
+
+{{/*
+Return the appropriate apiVersion for ingress object.
+*/}}
+{{- define "capabilities.ingress.apiVersion" -}}
+{{- if semverCompare "<1.14-0" ( .Values.kubeVersion | default .Capabilities.KubeVersion.Version ) -}}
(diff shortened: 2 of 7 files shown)


pierluigilenoci:
LGTM

pierluigilenoci (dismissed):

pierluigilenoci:
After approving, I had second thoughts. 😞 

The new `kubeVersion` value is used only within the PDB but `.Capabilities` is used elsewhere in the chart.
This option should override in all `.Capabilities` use cases. 

@desaintmartin what do you think about it?

julien-sugg:
Indeed, I only focused on the PDB on purpose as I didn't have much time to work on it.

I can attempt to perform the fix everywhere else by the next week if you desire to achieve consistency, which totally makes sense IMO.

I believe it would also be a good opportunity to also rework/harmonize the labels as it seems that some inconsistencies arose when labels were refactored to meet [Kubernetes Labels Recommendations](https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/).

pierluigilenoci:
Adding an inconsistent option doesn't make sense to me because it can mislead someone in the configuration.

Obviously if in the meantime we manage to align ourselves with the recommendations even better!

For now, thanks for your work.

julien-sugg:
I got some time to harmonize the changes around `.Capabilities` and refactored a bit the Ingress in the meanwhile to provide better support around changes from Kubernetes 1.18 (`pathType` and `ingressClassName`) as the `pathType` field doesn't necessarily implies the use of `networking.k8s.io/v1`, and also provided some refacts around `backend` field via a template.

pierluigilenoci:
Now looks better to me.

@desaintmartin could you please also take a look?

pierluigilenoci (approved):
LGTM, let's wait another reviewer opinion

desaintmartin (approved):
As a remark, I would forbid the use of ingressclass if not supported, but that's probably outside of the scope of the PR and can be done later.

LGTM for me as well!

PR #68, opened 2021-11-15
Feature: Add support for defining custom container securityContext

### LINKED ISSUE #67: Add support for custom Container SecurityContext
Problem: 
Currently you can't define a custom `securityContext` for a container.
This is a problem because you can't set for example `allowPrivilegeEscalation` or `capabilities` to the pod. ([Documentation](https://kubernetes.io/docs/tasks/configure-pod-container/security-context/))
https://github.com/oauth2-proxy/manifests/blob/7089e0ec39ed44dc27e0df25c23dd15d1b6a06b9/helm/oauth2-proxy/templates/deployment.yaml#L191-L194

Solution:
Backwards compatible:
```yaml
{{- if .Values.securityContext.enabled }}
{{- $securityContext := unset .Values.securityContext "enabled" }}
        securityContext:
          {{- toYaml $securityContext | nindent 10 }}
{{- end }}
```

Or if we don't want backwards compatibility do the same as with podSecurityContext:
https://github.com/oauth2-proxy/manifests/blob/7089e0ec39ed44dc27e0df25c23dd15d1b6a06b9/helm/oauth2-proxy/templates/deployment.yaml#L37-L40

Which do you prefer? I could submit a pull request after discussion

rlacko58:
closes #67

### DIFF
--- a/helm/oauth2-proxy/Chart.yaml
+++ b/helm/oauth2-proxy/Chart.yaml
@@ -1,5 +1,5 @@
 name: oauth2-proxy
-version: 5.1.0
+version: 5.2.0
 apiVersion: v2
 appVersion: 7.2.0
 home: https://oauth2-proxy.github.io/oauth2-proxy/
--- a/helm/oauth2-proxy/templates/deployment.yaml
+++ b/helm/oauth2-proxy/templates/deployment.yaml
@@ -189,9 +189,10 @@ spec:
 {{ toYaml .Values.extraVolumeMounts | indent 8 }}
 {{- end }}
 {{- if .Values.securityContext.enabled }}
+{{- $securityContext := unset .Values.securityContext "enabled" }}
         securityContext:
-          runAsNonRoot: {{ .Values.securityContext.runAsNonRoot }}
-{{- end}}
+          {{- toYaml $securityContext | nindent 10 }}
+{{- end }}
       volumes:
 {{- with .Values.config.google }}
 {{- if and .adminEmail (or .serviceAccountJson .existingSecret) }}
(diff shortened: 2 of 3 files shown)


pierluigilenoci:
This PR has some problems:
* chart version bump is missing
* in this way `enabled: true` also ends up inside the `securityContext`

It needs to be fixed up a bit and will probably introduce some breaking changes.

rlacko58:
@pierluigilenoci `enabled: true` won't end up in the `securityContext`.
It is removed with Helm's unset function in line 192:
```yaml
{{- $securityContext := unset .Values.securityContext "enabled" }}
```

Or should I introduce some breaking changes like this?
```yaml
 {{- with .Values.podSecurityContext }} 
 securityContext: 
   {{- toYaml . | nindent 8 }} 
 {{- end }} 
```

pierluigilenoci:
> @pierluigilenoci `enabled: true` won't end up in the `securityContext`. It is removed with Helm's unset function in line 192:

You're right, I didn't notice.

However, the version still needs to be updated because we merged another PR a few minutes ago.

For me it is enough but I would like the opinion of another maintainer.

rlacko58:
Alright, I will update the bump

pierluigilenoci (dismissed):

desaintmartin (changes requested):
Sorry for the huge delay.
Could you rebase (it should be easy) and bump a minor version of the chart (according to semver, this is new feature) instead of patch version?

rlacko58 (commented):
I've bumped it to the next minor version

desaintmartin (approved):

PR #36, opened 2021-06-07
add ability to specify alpha config file

gitdr:

PR #71, opened 2021-11-26
add support for alpha configuration.

UiP9AV6Y:
the alpha configuration provides access to experimental, as well
as advanced features not available with the "regular" config.

this change add support for either using an existing configmap
or creating one using the provided values.

obsoletes #36

### DIFF
--- a/helm/oauth2-proxy/templates/configmap-alpha.yaml
+++ b/helm/oauth2-proxy/templates/configmap-alpha.yaml
@@ -1,0 +1,32 @@
+{{- if .Values.alphaConfig.enabled }}
+{{- if not .Values.alphaConfig.existingConfig }}
+apiVersion: v1
+kind: ConfigMap
+metadata:
+{{- if .Values.alphaConfig.annotations }}
+  annotations: {{- toYaml .Values.alphaConfig.annotations | nindent 4 }}
+{{- end }}
+  labels:
+    app: {{ template "oauth2-proxy.name" . }}
+    {{- include "oauth2-proxy.labels" . | indent 4 }}
+  name: {{ template "oauth2-proxy.fullname" . }}-alpha
+data:
+  oauth2_proxy.yml: |
+    ---
+    server:
+      BindAddress: '0.0.0.0:4180'
+    {{- if .Values.alphaConfig.serverConfigData }}
+    {{- toYaml .Values.alphaConfig.serverConfigData | nindent 6 }}
+    {{- end }}
+    {{- if .Values.metrics.enabled }}
+    metricsServer:
+      BindAddress: '0.0.0.0:44180'
+    {{- if .Values.alphaConfig.metricsConfigData }}
+    {{- toYaml .Values.alphaConfig.metricsConfigData | nindent 6 }}
+    {{- end }}
+    {{- end }}
+    {{- if .Values.alphaConfig.configData }}
+    {{- toYaml .Values.alphaConfig.configData | nindent 4 }}
+    {{- end }}
+{{- end }}
+{{- end }}
--- a/helm/oauth2-proxy/templates/deployment.yaml
+++ b/helm/oauth2-proxy/templates/deployment.yaml
@@ -54,9 +54,13 @@ spec:
         image: "{{ .Values.image.repository }}:{{ .Values.image.tag }}"
         imagePullPolicy: {{ .Values.image.pullPolicy }}
         args:
+        {{- if .Values.alphaConfig.enabled }}
+          - --alpha-config=/etc/oauth2_proxy/oauth2_proxy.yml
+        {{- else }}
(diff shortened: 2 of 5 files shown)


pierluigilenoci:
@UiP9AV6Y could you please solve the conflict?

helm/oauth2-proxy/templates/configmap-alpha.yaml
```diff
@@ -0,0 +1,43 @@
...
+    ---
+    server:
+      BindAddress: '0.0.0.0:4180'
+    {{- if .Values.metrics.enabled }}
+    metricsServer:
+      BindAddress: '0.0.0.0:44180'
+    {{- end }}
+    {{- if .Values.alphaConfig.providers }}
+    providers:
+      {{- toYaml .Values.alphaConfig.providers | indent 6 }}
```
myukselen, line 25: `nindent` might be used instead of `indent`

myukselen: same as for line 29, 33, and 37

helm/oauth2-proxy/templates/deployment.yaml
```diff
@@ -44,10 +44,14 @@ spec:
         image: "{{ .Values.image.repository }}:{{ .Values.image.tag }}"
         imagePullPolicy: {{ .Values.image.pullPolicy }}
         args:
+        {{- if or .Values.alphaConfig.enabled }}
```
morganchristiansson, line 47:
```suggestion
        {{- if .Values.alphaConfig.enabled }}
```

UiP9AV6Y:
i have also added a chart testing values file to have linter coverage for this feature as well

helm/oauth2-proxy/templates/configmap-alpha.yaml
```diff
@@ -0,0 +1,39 @@
...
+    upstreamConfig: {{ toYaml .Values.alphaConfig.upstreamConfig | nindent 6 }}
+    {{- end }}
+    {{- if .Values.alphaConfig.injectRequestHeaders }}
+    injectRequestHeaders: {{ toYaml .Values.alphaConfig.injectRequestHeaders | nindent 6 }}
+    {{- end }}
+    {{- if .Values.alphaConfig.injectResponseHeaders }}
+    injectResponseHeaders: {{- toYaml .Values.alphaConfig.injectResponseHeaders | nindent 6 }}
+    {{- end }}
+    {{- if .Values.alphaConfig.configData }}
+    {{- toYaml .Values.alphaConfig.configData | nindent 4 }}
```
desaintmartin, line 36: Should we just keep this one? What is the added values of the previous ones?

pierluigilenoci (approved):

pierluigilenoci:
@UiP9AV6Y could you please do a version bump?

desaintmartin (commented):
Sorry for the huge delay and thanks for the PR!
See my small note.

helm/oauth2-proxy/templates/configmap-alpha.yaml
UiP9AV6Y: the `configData` provides a future-proof solution, for 
changes to the oauth2-proxy config format which are not 
(yet) supported by the chart.

the other keys represent the currently supported 
configuration values and are provided to allow easier 
patching /chart-deployment. yes, in theory i could simply 
use `configData` as the only configuration input, but then
you would have to provide the entire config as a single 
blob (not really an issue, when you configure the chart 
deployment via files, but get cumbersome when you 
deploy it e.g. via CLI)

i can change the implementation, if you prefer to have the 
entire config under a single key

UiP9AV6Y: i have reworked the PR to make no 
assumptions about the alpha config 
structure apart from the `server` and
`metricsServer` keys (which can be
enrichted using separate values)

desaintmartin:
I see that tests are failing :(

pierluigilenoci:
@UiP9AV6Y could you please fix the PR?

UiP9AV6Y:
i fixed all the issues (hopefully) and rebased the changes

pierluigilenoci:
@UiP9AV6Y still failing.

UiP9AV6Y:
i have remove the chart-testing values file for this change, b/c oauth2-proxy requires a valid provider configuration. it checks the connectivity on startup and terminates otherwise.

pierluigilenoci:

desaintmartin (approved):
Thanks!

desaintmartin:
> @desaintmartin could you please take a look?

Hehe I did not even see your comment when merging!

pierluigilenoci:
@desaintmartin I was very surprised by the speed, in fact. 😝

PR #72, opened 2021-11-30
Fix ServiceMonitor selector

LittleFox94:
The ServiceMonitor selector uses hardcoded labels not matching the ones set by the Chart, this PR fixes that by using the appropriate helper for that.

Also updates the ServiceMonitor labels to use the common helper, so it's labels match the labels on the other resources.

### DIFF
--- a/helm/oauth2-proxy/templates/servicemonitor.yaml
+++ b/helm/oauth2-proxy/templates/servicemonitor.yaml
@@ -19,8 +19,7 @@ spec:
   jobLabel: {{ template "oauth2-proxy.fullname" . }}
   selector:
     matchLabels:
-      app: {{ template "oauth2-proxy.name" . }}
-      release: {{ .Release.Name }}
+      {{- include "oauth2-proxy.selectorLabels" . | indent 6 }}
   namespaceSelector:
     matchNames:
       - {{ .Release.Namespace }}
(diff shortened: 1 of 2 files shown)


helm/oauth2-proxy/templates/servicemonitor.yaml
```diff
@@ -9,20 +9,18 @@ metadata:
   namespace: {{ .Release.Namespace | quote }}
 {{- end }}
   labels:
-    app: {{ template "oauth2-proxy.name" . }}
-    chart: {{ template "oauth2-proxy.chart" . }}
-    release: {{ .Release.Name }}
-    heritage: {{ .Release.Service }}
+  labels:
```
desaintmartin, line 12: Duplicate?

helm/oauth2-proxy/templates/servicemonitor.yaml
```diff
@@ -9,20 +9,18 @@ metadata:
...
     prometheus: {{ .Values.metrics.servicemonitor.prometheusInstance }}
+    app: {{ template "oauth2-proxy.name" . }}
```
desaintmartin, line 14: Is this one needed?

desaintmartin (commented):
thanks!

helm/oauth2-proxy/templates/servicemonitor.yaml
LittleFox94: indeed, thanks

helm/oauth2-proxy/templates/servicemonitor.yaml
LittleFox94: this same additional label is on the service, too - kept it for consistency

desaintmartin (approved):

PR #74, opened 2021-12-28
Add option to set dnsPolicy on pod.

ebuzzz:


pierluigilenoci (changes requested):
A few things are missing from this PR: 
* Chart version bump
* Default value in `values.yaml` 
* Configuration details inside the README.md 

And maybe something else that I didn't notice

pierluigilenoci:
@ebuzzz is this feature still needed?

pierluigilenoci:
@ebuzzz so?

ebuzzz:
@pierluigilenoci I've changed my intern DNS structure, so this change is no longer needed for my specific use-case.

PR #76, opened 2022-01-18
Added ability to deploy extra K8s manifests

nxf5025:
Signed-off-by: Nick Fisher <nxf5025@gmail.com>

By adding a secretProviderClass object to the helm chart we are able to mount secrets (such as our client secret) on the fly by leveraging the Secrets Store CSI Driver.

Documentation can be found here: https://github.com/kubernetes-sigs/secrets-store-csi-driver

### DIFF
--- a/helm/oauth2-proxy/Chart.yaml
+++ b/helm/oauth2-proxy/Chart.yaml
@@ -1,5 +1,5 @@
 name: oauth2-proxy
-version: 5.0.6
+version: 5.1.0
 apiVersion: v2
 appVersion: 7.2.0
 home: https://oauth2-proxy.github.io/oauth2-proxy/
--- a/helm/oauth2-proxy/README.md
+++ b/helm/oauth2-proxy/README.md
@@ -170,7 +170,8 @@ Parameter | Description | Default
 `metrics.servicemonitor.interval` | Prometheus scrape interval | `60s`
 `metrics.servicemonitor.scrapeTimeout` | Prometheus scrape timeout | `30s`
 `metrics.servicemonitor.labels` | Add custom labels to the ServiceMonitor resource| `{}`
-
+`extraObjects` | Extra K8s manifests to deploy | `[]`
+ 
 Specify each parameter using the `--set key=value[,key=value]` argument to `helm install`. For example,
 
 ```console
--- a/helm/oauth2-proxy/templates/extra-manifests.yaml
+++ b/helm/oauth2-proxy/templates/extra-manifests.yaml
@@ -1,0 +1,4 @@
+{{ range .Values.extraObjects }}
+---
+{{ tpl (toYaml .) $ }}
+{{ end }}
--- a/helm/oauth2-proxy/values.yaml
+++ b/helm/oauth2-proxy/values.yaml
@@ -265,3 +265,33 @@ metrics:
     scrapeTimeout: 30s
     # Add custom labels to the ServiceMonitor resource
     labels: {}
+
+# Extra K8s manifests to deploy
+extraObjects: []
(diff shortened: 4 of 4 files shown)


helm/oauth2-proxy/templates/secret-provider-class.yaml
```diff
@@ -0,0 +1,12 @@
+{{- if .Values.secretProviderClass }}
```
pierluigilenoci, line 1: To me make sense to use the same approach of other k8s object
Ref: https://github.com/oauth2-proxy/manifests/blob/fb57298898d94fc3491e0b4c6fa695a63d6f2111/helm/oauth2-proxy/values.yaml#L104

helm/oauth2-proxy/Chart.yaml
```diff
@@ -1,5 +1,5 @@
 name: oauth2-proxy
-version: 5.0.6
```
pierluigilenoci, line 2: Is this a breaking change?

helm/oauth2-proxy/templates/secret-provider-class.yaml
nxf5025: Reason I didn't go that route is if content is specified in the secretProviderClass then it can be assumed that it's enabled. This is similar to what is done in the kube-prometheus-stack helm chart - https://github.com/prometheus-community/helm-charts/blob/main/charts/kube-prometheus-stack/templates/prometheus/csi-secret.yaml

helm/oauth2-proxy/Chart.yaml
nxf5025: This is not a breaking change hence the minor version being bumped. This is only adding functionality in a backwards compatible manner.

pierluigilenoci:

desaintmartin:
I am not sure to understand the link between this and oauth2-proxy. Shouldn't it be managed outside of this chart in order to keep it simple?

For example, in my organization, we manage such things (in fact, RBACs, but it works the same) in a dedicated chart / release.

nxf5025:
@desaintmartin It actually does make sense for this to be managed directly with the chart since if the CSI driver is being used then this chart won't work without the associated secretProviderClass.

However check out this discussion on a similar PR for the argo-cd helm chart: https://github.com/argoproj/argo-helm/pull/1094#issuecomment-1016724994

I'm completely fine with switching up the logic to match what we did over there. Just let me know which route you want me to take.

nxf5025:
@desaintmartin / @pierluigilenoci - Any input on if you want me to go the more generic route?

pierluigilenoci:
@nxf5025 among the maintainers of the chart I am the latest arrival and I try to lend a hand above all to rough out the PR and answer the more technical questions. I would never allow myself to try to impose my views on "structural chart changes" so don't make my opinion binding. I refer you to the opinion of those with better know-how.

But if you ask me for my opinion, the ["bitnami" option](https://github.com/nxf5025/argo-helm/blob/03a9908cb2bb8cc97fc5f16c3221dfcc5c2020ae/charts/argo-cd/values.yaml#L62) to allow extra-manifest seems like the way to go. The stability of the chart is not disturbed and the possibility is offered to others to attach whatever they want to the installation.

@desaintmartin what do you think?

nxf5025:
@pierluigilenoci - After discussion on a few other the other projects where I've opened similar PR's I fully agree. I just updated this to support the more general extra manifests.

pierluigilenoci:
@desaintmartin I think it's better now.

@nxf5025 thank you!

pierluigilenoci (approved):

desaintmartin (approved):
Sorry for the lag.
This is quite generic indeed and could be useful for a lot of different uses!
lgtm

PR #81, opened 2022-02-15
fix redis password env ref when using existingSecret

### LINKED ISSUE #78: redis.existingSecret doesn't take effect when redis.password is set to empty string
Can't use an existingSecret for redis session storage in order to set OAUTH2_PROXY_REDIS_PASSWORD when the password has a empty string assigned.

Doesn't work:
```yaml
# values.yaml
...
sessionStorage:
  type: redis
  redis:
    existingSecret: oauth2-proxy-redis-access
    password: ""
...
```

Works:
```yaml
# values.yaml
...
sessionStorage:
  type: redis
  redis:
    existingSecret: oauth2-proxy-redis-access
    password: someValue
...
```

Fix suggestion:
```yaml
# templates/deployment.yaml
...
        {{- if or .Values.sessionStorage.redis.password .Values.sessionStorage.redis.existingSecret }}
        - name: OAUTH2_PROXY_REDIS_PASSWORD
          valueFrom:
            secretKeyRef:
              name: {{ if .Values.sessionStorage.redis.existingSecret }} {{ .Values.sessionStorage.redis.existingSecret }}{{ else }} {{ template "oauth2-proxy.fullname" . }}-redis-access{{ end }}
              key: redis-password
        {{- end }}
...
```

bluebrown:
There is a bug that leads to a situation where the environment variable OAUTH2_PROXY_REDIS_PASSWORD is not set on the deployment pod when using an existingSecret.

Fixes #78

### DIFF
--- a/helm/oauth2-proxy/templates/deployment.yaml
+++ b/helm/oauth2-proxy/templates/deployment.yaml
@@ -6,7 +6,7 @@ metadata:
 {{- include "oauth2-proxy.labels" . | indent 4 }}
   {{- if .Values.deploymentAnnotations }}
   annotations:
-{{ toYaml .Values.deploymentAnnotations | indent 8 }} 
+{{ toYaml .Values.deploymentAnnotations | indent 8 }}
   {{- end }}
   name: {{ template "oauth2-proxy.fullname" . }}
 spec:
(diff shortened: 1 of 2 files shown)


helm/oauth2-proxy/templates/deployment.yaml
```diff
@@ -6,7 +6,7 @@ metadata:
 {{- include "oauth2-proxy.labels" . | indent 4 }}
   {{- if .Values.deploymentAnnotations }}
   annotations:
-{{ toYaml .Values.deploymentAnnotations | indent 8 }} 
+{{ toYaml .Values.deploymentAnnotations | indent 8 }}
```
bluebrown, line 9: I just noticed this. Could be from auto formatting. Is thos ok?

desaintmartin (approved):
Thanks!

Issue #82, opened 2022-02-15
redis standalone url is not correctly computed


bluebrown:
Currently, the redis url is computed based on the helper `oauth2-proxy.fullname`. This works only if the release name is the chart name.  Actually, redis does not respect this helper. It will compute its name based on `fullnameOverride`, `nameOverride` and `.Release.Name`, as it can be seen [here](https://github.com/bitnami/charts/blob/41abb9a7e89bc3db0df7064348fda7eff0c408a4/bitnami/common/templates/_names.tpl#L21).

Additionally, the service port can be set via `master.service.ports.redis` which is not taken into consideration in the oauth2-proxy chart.

PR #83, opened 2022-02-15
compute the redis standalone url correctly

bluebrown:
compute the redis url based on release name or name overrides instead of oauth2-proxy fullname helper

Fixed: #82

### DIFF
--- a/helm/oauth2-proxy/templates/_helpers.tpl
+++ b/helm/oauth2-proxy/templates/_helpers.tpl
@@ -78,11 +78,21 @@ Create the name of the service account to use
 {{- end -}}
 {{- end -}}
 
-
+{{/*
+Compute the redis url if not set explicitly.
+*/}}
 {{- define "oauth2-proxy.redisStandaloneUrl" -}}
+{{- $masterPort := .Values.redis.master.service.ports.redis -}}
 {{- if .Values.sessionStorage.redis.standalone.connectionUrl -}}
 {{ .Values.sessionStorage.redis.standalone.connectionUrl }}
+{{- else if .Values.redis.fullnameOverride -}}
+{{- printf "redis://%s-master:%.0f" (.Values.redis.fullnameOverride | trunc 63 | trimSuffix "-") $masterPort -}}
 {{- else -}}
-{{- printf "redis://%s-redis-master:6379" (include "oauth2-proxy.fullname" .) -}}
+{{- $name := default "redis" .Values.redis.nameOverride -}}
+{{- if contains $name .Release.Name -}}
+{{- printf "redis://%s-master:%.0f" (.Release.Name | trunc 63 | trimSuffix "-") $masterPort -}}
+{{- else -}}
+{{- printf "redis://%s-master:%.0f" (printf "%s-%s" .Release.Name $name | trunc 63 | trimSuffix "-") $masterPort -}}
+{{- end -}}
 {{- end -}}
 {{- end -}}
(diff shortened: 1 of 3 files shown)


pierluigilenoci (approved):

pierluigilenoci:

pierluigilenoci:
@bluebrown why did you close the PR?
Did I miss something?

bluebrown:
Ah snap, this was a mistake. I deleted the branch. But I still have the changes. I could solve this with the other PR I am going to open now. It would be a bit messy, but they build on top of each other anyway.

bluebrown:
@pierluigilenoci, I have restored the branch. So they can be now merged one after the other. Although, as I said, #84 is kind of building on top of it.

helm/oauth2-proxy/templates/_helpers.tpl
```diff
@@ -78,11 +78,21 @@ Create the name of the service account to use
...
-
+{{/*
+Compute the redis url if not set explicitly.
+*/}}
 {{- define "oauth2-proxy.redisStandaloneUrl" -}}
+{{- $masterPort := .Values.redis.master.service.ports.redis -}}
 {{- if .Values.sessionStorage.redis.standalone.connectionUrl -}}
 {{ .Values.sessionStorage.redis.standalone.connectionUrl }}
+{{- else if .Values.redis.fullnameOverride -}}
+{{- printf "redis://%s-master:%.0f" (.Values.redis.fullnameOverride | trunc 63 | trimSuffix "-") $masterPort -}}
```
desaintmartin, line 89: You could probably build upon a mix of:
- How it is built in redis chart (using `common.names.fullname`)
- `common.names.dependency.fullname`

Like `printf "%s-master" (include "common.names.dependency.fullname" (dict "chartName" "redis" "chartValues" .Values.redis "context" $)`

in order not to have this whole logic hardcoded here.

bluebrown: the problem is that the helper anicipates as context the chart itself. I didnt find out how to pass in the redis chart instead of the oauth2-proxy chart. Because using the parent chart is not correct.  The chart name that is used in this helper should be *redis*.

bluebrown: Or is this the solution to the mentioned problem?

`dict "chartName" "redis" "chartValues" .Values.redis "context" $`

Because that would be amazing, that's right.

bluebrown: I think I get it it, you are passing in a dict as context that *mocks* the redis chart. It contains the required values for the helper to function. Thats cool. I will make the required changes in a bit. 

Thanks for the feedback.

helm/oauth2-proxy/templates/_helpers.tpl
```diff
@@ -78,11 +78,26 @@ Create the name of the service account to use
 {{- end -}}
 {{- end -}}
 
+{{/*
+Redis subcharts fullname
+*/}}
+{{- define "oauth2-proxy.redis.fullname" -}}
+{{- if .Values.redis.enabled -}}
+{{- include "common.names.fullname" (dict "Chart" (dict "Name" "redis") "Release" .Release "Values" .Values.redis) -}}
```
bluebrown, line 86: @desaintmartin, this is the best I could figure out. What do you think?

desaintmartin (dismissed):
Looks good! https://github.com/oauth2-proxy/manifests/runs/5222114748?check_suite_focus=true#step:9:926 shows that URL is correctly given to the Deployment.

desaintmartin (approved):

pierluigilenoci (approved):

Issue #84, opened 2022-02-16
use redis subchart pwd secret if no existing secret or password was set


bluebrown:
It would be nice that when the redis subchart is enabled and has auth enabled, the password would be read from the secret that is created by the redis subchart. 

This would remove the need to set the password twice or reference the secret which name is dependent on the release name. It would make the integreation with redis more straight forward and should work out of the box for most use cases.

If implemented, the redis integration via subchart would require only the below values to work. 

```yaml
sessionStorage:
  type: redis
redis:
  enabled: true
  auth:
    password: foobar
```

pierluigilenoci:
@bluebrown feel free to open a PR.

bluebrown:
3 pull requests. Im such a noob :(

bluebrown:
Next time I will just merge, then you have multiple commits. I tried squashing and force pushing, but it actually doesn't resolve merge conflicts, unlike a merge commit.

I would suggest that you squash merge the PR afterwards.

pierluigilenoci:
From my point of view, the important thing is goodwill and the final result. 
I don't know what other people think but that's okay with me.
The next time will be better.

PR #87, opened 2022-02-21
use password secret from redis subchart

bluebrown:

Fixes : #84

### DIFF
--- a/.github/workflows/lint-test.yaml
+++ b/.github/workflows/lint-test.yaml
@@ -41,5 +41,9 @@ jobs:
         uses: helm/kind-action@v1.2.0
         if: steps.list-changed.outputs.changed == 'true'
 
+      - name: Install Prometheus Operator CRDs
+        id: prom
+        run: kubectl apply --server-side -f https://raw.githubusercontent.com/prometheus-operator/prometheus-operator/v0.54.0/example/prometheus-operator-crd/monitoring.coreos.com_servicemonitors.yaml
+
       - name: Run chart-testing (install)
         run: ct install --config ct.yaml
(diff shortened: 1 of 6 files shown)


bluebrown:
I am not sure why the tests here fail. I have tested this locally. The chart works as is. It seems the test is looking for the secret, which is no longer created.

```bash
Warning  Failed     7m48s (x12 over 10m)  kubelet            Error: secret "oauth2-proxy-06u7ma6raq-redis-access" not found
```

I have added this to the secret, which can lead to a situation where it's not created. But in this case it should actually not get created, as far as I can tell. Maybe we need to adjust the tests for this.

```yaml
{{- with .Values.sessionStorage }}
{{- if and (eq .type "redis") (not .redis.existingSecret) (or .redis.password .redis.sentinel.password) }}
# create secret
```

bluebrown:
I did not fully understand how the tests are created and run. I could add some additional test cases. Furthermore, I don't think this PR is actually covered, since it's setting the PW explicitly.

pierluigilenoci:
@bluebrown this PR needs a rebase.

bluebrown:
idk, I suck at git. Now its broken again.

bluebrown:
It would be too bad if you did a squash commit when merging. Like this I try to rebase it in a single commit on my end but the confilcts keep staying.

bluebrown:
Now its 2 commits. I didnt know how to do it better. Sorry.

bluebrown:
I will do it again from master. I don't know, this rebase stuff is weird. Furthermore, I prefer a regular merge.

PR #89, opened 2022-02-22
Add logic for customising service.targetPort

endakelly:
Update service.yaml template to enable using a custom targetPort not tied to the httpScheme﻿


pierluigilenoci:
@endakelly a few things are missing in this PR.
Certainly the bump of the chart version.
I am also not too convinced that the approach used makes sense.
Rather it makes sense for me to put the containerPort value inside the values file and allow customization from there. 

What do you think?

endakelly:
> @endakelly a few things are missing in this PR. Certainly the bump of the chart version. I am also not too convinced that the approach used makes sense. Rather it makes sense for me to put the containerPort value inside the values file and allow customization from there.
> 
> What do you think?

Makes sense. I've made an update if you want to have a look

helm/oauth2-proxy/templates/deployment.yaml
```diff
@@ -147,7 +147,13 @@ spec:
 {{ tpl (toYaml .Values.extraEnv) . | indent 8 }}
         {{- end }}
         ports:
+        {{- if (and (eq .Values.httpScheme "http") (empty .Values.containerPort)) }}
           - containerPort: 4180
+        {{- else if (and (eq .Values.httpScheme "https") (empty .Values.containerPort)) }}
+          - containerPort: 4443
```
bluebrown, line 153: Its by default 443

<html>
<body>


--https-address | string | <addr>:<port> to listen on for HTTPS clients | ":443"
-- | -- | -- | --



</body>
</html>

https://oauth2-proxy.github.io/oauth2-proxy/docs/configuration/overview

helm/oauth2-proxy/templates/deployment.yaml
```diff
@@ -147,7 +147,13 @@ spec:
...
+        {{- else }}
+          - containerPort: {{ .Values.containerPort }}
```
bluebrown, line 155: Container port should take precedence, I think you cannot have the schema empty, so this else block will never be used.

endakelly: I've changed the order around so containerPort takes precedence. 

This isn't assuming the httpScheme is empty. The logic will set the value of the containerPort to a default value if the httpScheme is set and the containerPort is not. If the containerPort is defined then it will be set to what the user entered in the values.yaml. They will continue to get an error if they don't configure the httpScheme value.

pierluigilenoci:

helm/oauth2-proxy/templates/deployment.yaml
bluebrown: @endakelly you are right. I didn't look properly. In that case, you could have left the order as is. Now the checks for empty containerPort are redundant, in the new commit.

helm/oauth2-proxy/templates/deployment.yaml
desaintmartin: Unfortunately, I believe this would not work if securityContext is enabled (not running as root). Should we do something here, or accept the defaults?

helm/oauth2-proxy/values.yaml
```diff
@@ -62,6 +62,11 @@ image:
 # imagePullSecrets:
   # - name: myRegistryKeySecretName
 
+# Set a custom containerPort if required.
+# This will default to 4180 if this value is not set and the httpScheme set to http
+# This will default to 443 if this value is not set and the httpScheme set to https
+# containerPort: 4180
```
desaintmartin, line 68: - Could you also document the value in the README?
- Suggestion/discussion: Another simpler way of self-documenting would be to introduce a `containerPortHttp` and `containerPortHttps` with sane defaults always defined.

helm/oauth2-proxy/Chart.yaml
```diff
@@ -1,5 +1,5 @@
 name: oauth2-proxy
-version: 6.1.2
+version: 6.1.3
```
desaintmartin, line 2:
```suggestion
version: 6.2.0
```

pierluigilenoci:
@endakelly could you please take a look?

endakelly:
Apologies for the delay in updates. I've updated the readme now. BTW, the issue I was having seemed to resolve itself after I started again and left as many variables and config values at their defaults. I'm not sure what was causing the issue for me - some combination of values maybe creating some sort of conflict.

pierluigilenoci:
@endakelly could you rebase?

helm/oauth2-proxy/values.yaml
pierluigilenoci: @desaintmartin is the `README.md` enough for now?

pierluigilenoci (dismissed):

pierluigilenoci (dismissed):

pierluigilenoci:

pierluigilenoci:

helm/oauth2-proxy/templates/deployment.yaml
desaintmartin: On another level, I'm afraid in the current form this will break existing installations (and thus is a breaking change). All releases using HTTPS will see the port on the Pod changing, but oauth2-proxy will still listen to the same old 4180 port.
I would suggest keeping it simple by just stating a port:
`- containerPort: {{ .Values.containerPort }}`
and a value, 4180 by default.

pierluigilenoci: @bluebrown can you please follow up on this?

desaintmartin: I think I mis-read the Service yaml. It is ok this way. I would just suggest changing default ports to 4443 and also override the defaults in an oauth2-proxy flag like it is done in https://github.com/oauth2-proxy/manifests/pull/117

pierluigilenoci:
@endakelly could you please rebase and fix your PR?

pierluigilenoci:
@endakelly could you please rebase and fix your PR?

pierluigilenoci:
@endakelly the bump of the chart version is missing.

pierluigilenoci (approved):

PR #90, opened 2022-02-22
Feature/ use redis secret from subchart

bluebrown:

Fixes: #84

### DIFF
--- a/helm/oauth2-proxy/templates/deployment.yaml
+++ b/helm/oauth2-proxy/templates/deployment.yaml
@@ -112,11 +112,17 @@ spec:
         {{- if eq (default "cookie" .Values.sessionStorage.type) "redis" }}
         - name: OAUTH2_PROXY_SESSION_STORE_TYPE
           value: "redis"
-        {{- if or .Values.sessionStorage.redis.password .Values.sessionStorage.redis.existingSecret }}
+         {{- if or .Values.sessionStorage.redis.existingSecret .Values.sessionStorage.redis.password (and .Values.redis.enabled (.Values.redis.auth).enabled )}}
         - name: OAUTH2_PROXY_REDIS_PASSWORD
           valueFrom:
             secretKeyRef:
-              name: {{ if .Values.sessionStorage.redis.existingSecret }} {{ .Values.sessionStorage.redis.existingSecret }}{{ else }} {{ template "oauth2-proxy.fullname" . }}-redis-access{{ end }}
+              {{- if .Values.sessionStorage.redis.existingSecret }}
+              name: {{ .Values.sessionStorage.redis.existingSecret }}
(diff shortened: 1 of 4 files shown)


pierluigilenoci (approved):
LGTM

pierluigilenoci:

helm/oauth2-proxy/templates/deployment.yaml
```diff
@@ -112,11 +112,17 @@ spec:
         {{- if eq (default "cookie" .Values.sessionStorage.type) "redis" }}
         - name: OAUTH2_PROXY_SESSION_STORE_TYPE
           value: "redis"
-        {{- if or .Values.sessionStorage.redis.password .Values.sessionStorage.redis.existingSecret }}
+         {{- if or .Values.sessionStorage.redis.existingSecret .Values.sessionStorage.redis.password (and .Values.redis.enabled (.Values.redis.auth).enabled )}}
```
desaintmartin, line 115:
```suggestion
        {{- if or .Values.sessionStorage.redis.password .Values.sessionStorage.redis.existingSecret (and .Values.redis.enabled .Values.redis.auth.enabled )}}
```

helm/oauth2-proxy/templates/deployment.yaml
```diff
@@ -112,11 +112,17 @@ spec:
...
         - name: OAUTH2_PROXY_REDIS_PASSWORD
           valueFrom:
             secretKeyRef:
-              name: {{ if .Values.sessionStorage.redis.existingSecret }} {{ .Values.sessionStorage.redis.existingSecret }}{{ else }} {{ template "oauth2-proxy.fullname" . }}-redis-access{{ end }}
+              {{- if .Values.sessionStorage.redis.existingSecret }}
```
desaintmartin, line 119: Shouldn't we do the same for sentinel? Note that even without it, this is still an improvement!

desaintmartin (commented):
Thanks! See my comments.

helm/oauth2-proxy/templates/deployment.yaml
bluebrown: I had it without the extra parenthesis as the beginning, but this is causing issues with Redis is not enabled. Helm errors on the evaluation of `.Values.redis.auth.enabled` for some reason, if the redis chart is not enabled. I would assume it stops early to evaluate when using the, and it's not the case, it seems.

helm/oauth2-proxy/templates/deployment.yaml
bluebrown: I guess yes, I didn't understand how this with sentinel works TBH. But I haven't really looked into that. I can take a look later and see if I can figure it out.

desaintmartin (approved):

PR #100, opened 2022-05-11
Bump oauth2-proxy image to latest release v7.2.1

andrewnicolalde:
Bug & security update which has no feature additions, per https://github.com/oauth2-proxy/oauth2-proxy/releases/tag/v7.2.1

PR #102, opened 2022-05-31
Bump oauth2-proxy image to latest release v7.3.0

andrewnicolalde:
Per the [release notes](https://github.com/oauth2-proxy/oauth2-proxy/releases/tag/v7.3.0), this release contains a number of security fixes and feature additions, but no breaking changes. The maintainers recommend updating to the latest version due to those security fixes "as soon as possible."

### DIFF
--- a/helm/oauth2-proxy/Chart.yaml
+++ b/helm/oauth2-proxy/Chart.yaml
@@ -1,7 +1,7 @@
 name: oauth2-proxy
-version: 6.2.1
+version: 6.2.2
 apiVersion: v2
-appVersion: 7.2.1
+appVersion: 7.3.0
 home: https://oauth2-proxy.github.io/oauth2-proxy/
 description: A reverse proxy that provides authentication with Google, Github or other providers
 keywords:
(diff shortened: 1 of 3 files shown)


andrewnicolalde:
This is quite similar to a previous PR https://github.com/oauth2-proxy/manifests/pull/100. Figured I'd leave this comment just in case someone confused a notification for that PR with this one, since the last one was reviewed so quickly!

jsargiot:
Do you think it makes sense to sync the chart version and the app version? I'm asking because it's somewhat confusing to install chart version 6.2.2 and get oauth2-proxy version 7.3.0.

I know it's not necessary the same thing (because you can override the app tag)... but just wondering if it makes sense

pierluigilenoci:
@jsargiot it makes no sense to me.

The two versions have two different meanings.
Many times new features are added to the chart without the software version changes.
How would you handle such a thing?
And what if a change in the chart is also a breaking change.

Sorry but, for me, it's a no-go across the board.

andrewnicolalde:
@pierluigilenoci 

Sounds good, but that isn't what this PR does :). It bumps the default oauth2-proxy image version to the latest version, but it only increases the chart version to `6.2.2`.

pierluigilenoci:
@andrewnicolalde so what?
If the version of the chart is not linked to the software version, a change of a minor version of the chart can also be linked to such an increase. As long as it's not a breaking change.

andrewnicolalde:
> @andrewnicolalde so what?
> 
> If the version of the chart is not linked to the software version, a change of a minor version of the chart can also be linked to such an increase. As long as it's not a breaking change.

I think we may be saying the same thing.. To be clear, I am not advocating syncing the helm chart and application release version numbers. I agree that it makes sense to keep those distinct.

pierluigilenoci:
> @pierluigilenoci I think we may be saying the same thing.

❤️

andrewnicolalde:
@pierluigilenoci Do you think this PR is ready to merge then? 🙂

pierluigilenoci (approved):

PR #116, opened 2022-10-25
Make password key in Kubernetes secret configurable

### LINKED ISSUE #115: Read redis sentinel password from same key as redis password
We're using the oauth2-proxy with a redis with activated sentinel. It's a default redis from the popular [bitnami redis chart](https://artifacthub.io/packages/helm/bitnami/redis).

To use the secret created by redis, we set `sessionStorage.redis.existingSecret` to the name of the existing secret. This is working for the redis password, because oauth2-proxy is expecting bitnami redis to provide a secret with the name `redis-password`, which it does: https://artifacthub.io/packages/helm/bitnami/redis?modal=template&template=secret.yaml

Sadly, it is not working for the redis sentinel password, because it is the same in bitnami redis and also called `redis-password`.

Is there a way to define the name of the password in the existing secret? Setting `sessionStorage.redis.sentinel.password` looks like the field I'm looking for, but is simply doing nothing.

Would it be an option to replace this line: https://github.com/oauth2-proxy/manifests/blob/b3b047c208d1053982154e1931cf2147d42a50c0/helm/oauth2-proxy/templates/deployment.yaml#L154

by
```
key: {{ .Values.sessionStorage.redis.sentinel.password | default "redis-sentinel-password" }}
```

We could simply use it setting `sessionStorage.redis.sentinel.password = "redis-password"`.

If you like this idea, I'll create a PR for it.

PS: we're using the following workaround in the mean time:

```
  extraEnv:
  - name: OAUTH2_PROXY_REDIS_SENTINEL_PASSWORD
    valueFrom:
      secretKeyRef:
        name: redis
        key: redis-password
```

alexbde:
Fixes #115 by adding an option to adjust the key of the secret data object containing the password for redis & its sentinel.

New configuration options:
- `sessionStorage.redis.passwordKey` (previously hard-coded as `redis-password`)
- `sessionStorage.redis.sentinel.passwordKey` (previously hard-coded as `redis-sentinel-password`)

I also added a new option `sessionStorage.redis.sentinel.existingSecret` to configure the sentinel indepent of the redis. This is not necessary to fix the issue above and can be removed. I just thought it might be useful for anybody in the future.

I also synchronized some documentation between README.md and values.yaml. Not sure if maintaining docs in two places is a good idea though. As a regular user of artifacthub I only have had a look at the values.yaml before creating this PR - just saying :)

### DIFF
--- a/helm/oauth2-proxy/Chart.yaml
+++ b/helm/oauth2-proxy/Chart.yaml
@@ -1,5 +1,5 @@
 name: oauth2-proxy
-version: 6.4.0
+version: 6.5.0
 apiVersion: v2
 appVersion: 7.3.0
 home: https://oauth2-proxy.github.io/oauth2-proxy/
--- a/helm/oauth2-proxy/README.md
+++ b/helm/oauth2-proxy/README.md
@@ -107,7 +107,7 @@ Parameter | Description | Default
 `alphaConfig.serverConfigData` | Arbitrary configuration data to append to the server section | `{}`
 `alphaConfig.metricsConfigData` | Arbitrary configuration data to append to the metrics section | `{}`
 `alphaConfig.configData` | Arbitrary configuration data to append | `{}`
-`alphaConfig.existingConfig` | existing Kubernetes configmap to use for the alpha configuration file. See [config template](https://github.com/oauth2-proxy/manifests/blob/master/helm/oauth2-proxy/templates/configmap-alpha.yaml) for the required values  | `nil`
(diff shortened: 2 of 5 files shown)


alexbde:
@pierluigilenoci any feedback? :)

desaintmartin (changes requested):
Thanks for the contribution! Could you bump minor version in chart.yaml since it is a new feature?

alexbde:
> Thanks for the contribution! Could you bump minor version in chart.yaml since it is a new feature?

Sure, done 👍

desaintmartin (approved):
Thanks!

pierluigilenoci:
@alexbde thank you a lot for your contribution.

PR #117, opened 2022-11-01
Set https port to 4443

Palollo:
Default port for https was 443 ([ref](https://github.com/oauth2-proxy/oauth2-proxy/blob/master/docs/docs/configuration/overview.md?plain=1#L126)) but it is not allowed for binding by a non-root user.
This is what occurs in the default image of this helm chart, so this is a fix for an error appeared when the value "httpScheme" is set to "https".

### DIFF
--- a/helm/oauth2-proxy/Chart.yaml
+++ b/helm/oauth2-proxy/Chart.yaml
@@ -1,5 +1,5 @@
 name: oauth2-proxy
-version: 6.3.0
+version: 6.3.1
 apiVersion: v2
 appVersion: 7.3.0
 home: https://oauth2-proxy.github.io/oauth2-proxy/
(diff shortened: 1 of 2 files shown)


desaintmartin:
Indeed. It is similar to https://github.com/oauth2-proxy/manifests/pull/117 but takes a more pragmatic approach. I 117 does not progress in a few days, let's merge this one.

desaintmartin (changes requested):
Could you bump version in chart.yaml?

Palollo:
> Indeed. It is similar to #117 but takes a more pragmatic approach. I 117 does not progress in a few days, let's merge this one.

I think you mean #89. I doubt that PR will work well: changing the container port is not enought for that he wants to. He forgot to change:
 - the app port in the "args" of the container (params "--http-address" and "--https-address") in deployment.yaml
 - and also the `targetPort` in the service.yaml file. 
If he is interested yet, he will see the mention here. I can add suggesting changes for clarification in his PR or include his changes in my PR with the needed additions.

> Could you bump version in chart.yaml?

Sure, it's done ;)

desaintmartin:
Let's merge this, it's easy to rebase in #89 when needed to make it configurable.

desaintmartin (approved):

PR #118, opened 2022-11-01
Optional nodePort selection for the service

Palollo:
Just adds the possibility to set the nodePort number.
Ref: https://kubernetes.io/docs/concepts/services-networking/service/#nodeport-custom-port

### DIFF
--- a/helm/oauth2-proxy/README.md
+++ b/helm/oauth2-proxy/README.md
@@ -159,6 +159,7 @@ Parameter | Description | Default
 `service.clusterIP` | cluster ip address | `nil`
 `service.loadBalancerIP` | ip of load balancer | `nil`
 `service.loadBalancerSourceRanges` | allowed source ranges in load balancer | `nil`
+`service.nodePortNumber` | external port number for the service when service.type is `NodePort` | `nil`
 `serviceAccount.enabled` | create a service account | `true`
 `serviceAccount.name` | the service account name | ``
 `serviceAccount.annotations` | (optional) annotations for the service account | `{}`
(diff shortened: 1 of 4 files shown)


Palollo:
The changes of this PR are tested in our cluster, setting the value of nodePortNumber and also leaving it empty.
The changes includes also the documentation update for the inclusion of the new properties.
Should I include the increment of chart version?

helm/oauth2-proxy/README.md
```diff
@@ -159,6 +159,7 @@ Parameter | Description | Default
 `service.clusterIP` | cluster ip address | `nil`
 `service.loadBalancerIP` | ip of load balancer | `nil`
 `service.loadBalancerSourceRanges` | allowed source ranges in load balancer | `nil`
+`service.nodePortNumber` | external port number for the service when service.type is `NodePort` | `nil`
```
desaintmartin, line 162: why `nodePortNumber` instead of `nodePort`?

desaintmartin (changes requested):
Could you bump minor version of Chart.yaml? This is new feature according to semver.

helm/oauth2-proxy/README.md
Palollo: Because there is already the property `service.portNumber` defined 5 lines above.

desaintmartin: Oh, that's ugly, actually! I would prefer to get rid of this one later on and introduce sane naming now, I think `nodePort` is both cleaner and closer to what the Helm community does.

Palollo: ok, done

desaintmartin:
Thanks!

PR #120, opened 2022-11-14
Correct label in NOTES.txt

philomory:
The `app` label on the pods is set based on the `name` template, not the `fullname` template. In cases where they are different (for example, if the release name is not `oauth2-proxy`, and neither `nameOverride` nor `fullnameOverride` are set), the instructions in the rendered notes would be incorrect.

### DIFF
--- a/helm/oauth2-proxy/Chart.yaml
+++ b/helm/oauth2-proxy/Chart.yaml
@@ -1,5 +1,5 @@
 name: oauth2-proxy
-version: 6.5.0
+version: 6.5.1
 apiVersion: v2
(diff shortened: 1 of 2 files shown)


pierluigilenoci:
@philomory a chart bump is needed.

philomory:
> @philomory a chart bump is needed.

@pierluigilenoci Done!

pierluigilenoci:

pierluigilenoci (approved):

desaintmartin:
Sorry, too late... approved... ;)

PR #126, opened 2022-12-20
fix(deps): upgrade redis subchart to 16.13.2.

desaintmartin:
Note that older redis versions are being stripped from bitnami repo: https://github.com/bitnami/charts/issues/8433 and prevent release (actually: preventing helm dep up)


desaintmartin:
Merging now since this breaks all CI.

PR #132, opened 2023-02-11
Pass oauth2-proxy config through helm templating

jkroepke:
This PR enabled passing configuration through the helm templating engine again.

Use cases:
We have a umbrella helm charts which bundles all helm charts together in one release. Using the pattern, we are able to establish a single configuration plane for multiple components.

In this case, we would like to share some configuration between oauth2-proxy and dex idp. For example dexidp and oauth-proxy shares the same issuer config, client id, client secret.

oauth2-proxy also shares some configuration with grafana, e.g. ingress hostname.

The [grafana](https://github.com/grafana/helm-charts/blob/2987944b43d77f9081fe7d7be40179e281df3917/charts/grafana/templates/ingress.yaml#L37) and [kube-prometheus-stack](https://github.com/prometheus-community/helm-charts/blob/edb39f574b95038229ca1b09468cc01b37214637/charts/kube-prometheus-stack/templates/alertmanager/secret.yaml#L19) already allows use similar configurations. 

Giving an example for values files

```yaml
oauth2-proxy:
  checkDeprecation: false
  config:
    clientSecret: '{{ $.Values.global.features.auth.secrets.proxyClientSecret }}'
    configFile: |
      oidc_issuer_url = "https://{{ $.Values.global.features.auth.hostname }}/dex"
  ingress:
    enabled: ture
    annotations:
      cert-manager.io/cluster-issuer: letsencrypt
    className: nginx
    path: /oauth2/
    pathType: Prefix
    hosts:
      - '{{ $.Values.global.features.grafana.hostname }}'
    tls:
      - secretName: tls-secret-grafana
        hosts:
          - '{{ $.Values.global.features.grafana.hostname }}'
```

### DIFF
--- a/helm/oauth2-proxy/Chart.yaml
+++ b/helm/oauth2-proxy/Chart.yaml
@@ -1,5 +1,5 @@
 name: oauth2-proxy
-version: 6.8.0
+version: 6.9.0
 apiVersion: v2
 appVersion: 7.4.0
 home: https://oauth2-proxy.github.io/oauth2-proxy/
--- a/helm/oauth2-proxy/ci/tpl-values.yaml
+++ b/helm/oauth2-proxy/ci/tpl-values.yaml
@@ -1,0 +1,21 @@
+extraEnv:
+  - name: TEST_ENV_VAR_2
+    value: '{{ $.Release.Name }}'
+ingress:
+  enabled: true
+  hosts:
+    - "{{ $.Release.Name }}.local"
+  tls:
+    - hosts:
+        - "{{ $.Release.Name }}.local"
+oauth2-proxy:
+  checkDeprecation: false
+  config:
+    clientSecret: '{{ $.Release.Name }}'
+    configFile: |
+      oidc_issuer_url = "https://{{ $.Release.Name }}/dex"
+
+pass_authorization_header: "true"
+
+extraArgs:
+  pass-authorization-header: "{{ $.Values.pass_authorization_header }}"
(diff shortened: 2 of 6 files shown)


pierluigilenoci (approved):
PR seems harmless enough to me so I'm in favor of it. 
But I would like a second opinion before doing the merge.

pierluigilenoci:
@jkroepke I'm sorry if I haven't been kind... I forgot to thank you for your contribution. 😛 
So thank you! 🚀 ❤️

pierluigilenoci:

desaintmartin (changes requested):
Note: we're using helmfile to do the same thing at my company (cons: not pure-helm, pros: separation of Releases allowing un-coupling but still have common configuration).

Could you just add/edit a test in the ci directory? This thing WILL definitely break in the future if there isn't. ;)

jkroepke:
@desaintmartin Tests added!

I agree about helmfile works better in that scenario, but looking forward to GitOps ArgoCD or flux, helmfile does not fit here anymore.

desaintmartin:
Hehe, the config file actually needs to be valid for the test to pass.

jkroepke:
Atleast the tests covers only, if the tpl function is working..

jkroepke:
@desaintmartin At least the tests covers only, if the tpl function is working..

desaintmartin (approved):
Thanks!

tim-jagenberg-piscada:
This created problems with boolean values:
https://github.com/oauth2-proxy/manifests/issues/133

jkroepke:
I will setup a pr in for fix this

PR #134, opened 2023-02-16
ensure values are string before passing to helm templating engine

### LINKED ISSUE #133: wrong type for value; expected string; got bool
Hello,
for setting>       `--set extraArgs.set-authorization-header="true" `
there is an error:
```
Error: template: oauth2-proxy/templates/deployment.yaml:76:32: executing "oauth2-proxy/templates/deployment.yaml" at <$value>: wrong type for value; expected string; got bool
```
for chart version 6.9.0 (6.8.0 works fine)

Have a nice day

jkroepke:
Introduced in #132 
Fixes #133

By using `--set extraArgs.set-authorization-header="true"`, helm will convert the value to a boolean value, since the value is boolean parsable.

For such scenarios, helm has additional flag `--set-string` to disable Interpretation of values.

The tpl function in helm always expecting strings. The PR ensure that the value is always a string before passing it into the tpl function.

### DIFF
--- a/helm/oauth2-proxy/Chart.yaml
+++ b/helm/oauth2-proxy/Chart.yaml
@@ -1,5 +1,5 @@
 name: oauth2-proxy
-version: 6.9.0
+version: 6.9.1
 apiVersion: v2
 appVersion: 7.4.0
 home: https://oauth2-proxy.github.io/oauth2-proxy/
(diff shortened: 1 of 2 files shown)


pierluigilenoci (approved):

pierluigilenoci:

desaintmartin (approved):
lgtm! Indeed the current implementation breaks simple use cases! Thanks!

PR #138, opened 2023-03-09
Add extraContainers: [] configuration option

illrill:
Add `extraContainers: []` configuration option to the chart, so that it's possible to insert a sidecar container. Could for example be useful to achieve setups like [this](https://www.callumpember.com/Kubernetes-A-Single-OAuth2-Proxy-For-Multiple-Ingresses/).

### DIFF
--- a/helm/oauth2-proxy/Chart.yaml
+++ b/helm/oauth2-proxy/Chart.yaml
@@ -1,5 +1,5 @@
 name: oauth2-proxy
-version: 6.9.2
+version: 6.10.0
 apiVersion: v2
 appVersion: 7.4.0
 home: https://oauth2-proxy.github.io/oauth2-proxy/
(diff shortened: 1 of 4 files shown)


helm/oauth2-proxy/Chart.yaml
```diff
@@ -1,5 +1,5 @@
 name: oauth2-proxy
-version: 6.9.1
+version: 6.9.2
```
pierluigilenoci, line 2: This is a new feature, so please bump the minor instead.

pierluigilenoci (changes requested):

pierluigilenoci:
@illrill, thank you for your contribution.

illrill:
> @illrill, thank you for your contribution.

Thanks for handling the version update! I accidentally removed the review request from @desaintmartin while I was trying to reset your request for change.

pierluigilenoci (approved):

PR #139, opened 2023-03-14
fix: allow literal 0 values in deployment.yaml extraArgs templating

detvdl:
Some of the available arguments to be passed to `extraArgs` allow for literal 0 values, such as the `--cookie-refresh` setting.

However, setting it to `0` causes the `{{ if $value }}` check to logically return as falsy, thanks to gotpl.

This replacement check instead looks for invalid value kinds (i.e. nil/null, explicitly empty)

### DIFF
--- a/helm/oauth2-proxy/Chart.yaml
+++ b/helm/oauth2-proxy/Chart.yaml
@@ -1,5 +1,5 @@
 name: oauth2-proxy
-version: 6.10.1
+version: 6.13.0
 apiVersion: v2
 appVersion: 7.4.0
 home: https://oauth2-proxy.github.io/oauth2-proxy/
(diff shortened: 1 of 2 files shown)


pierluigilenoci:
@detvdl could you please add also a chart version bump? And maybe a test?

detvdl:
Hey @pierluigilenoci , sorry for the lack of response, totally lost track of this. Will take a look at adding a test later today.

pierluigilenoci:
@detvdl, thank you!

detvdl:
Hey @pierluigilenoci ,

Given the limitations/scope of chart-testing (CT), I don't see an ideal way to test whether literal 0 values are in fact passed through to the deployment correctly.
I could just add something like
```yaml
extraArgs:
  cookie-refresh: 0
```
To any of the test configurations, but that won't verify anything, as the problem was that values simply do not get passed due to the conditional, and that won't crash the container(s) nor get caught by linting

Which approach do you prefer for testing:

1. Helm chart tests (https://helm.sh/docs/topics/chart_tests/) which invoke an actual container to verify whether the current template passes these value types
2. Custom tests with e.g. [TerraTest](https://terratest.gruntwork.io/) in Go, which purely act on the rendered templates
3. Other?

pierluigilenoci (approved):

detlevvandaele-tomtom:
No time wasted, no worries 🙂

PR #140, opened 2023-03-28
Way to disable appProtocol

none0nfg:
[K8s older than 1.20](https://kubernetes.io/docs/concepts/services-networking/service/#application-protocol), can't understand appProtocol in service description, so let's add way to disable appProtocol.

### DIFF
--- a/helm/oauth2-proxy/Chart.yaml
+++ b/helm/oauth2-proxy/Chart.yaml
@@ -1,5 +1,5 @@
 name: oauth2-proxy
-version: 6.13.3
+version: 6.14.1
 apiVersion: v2
 appVersion: 7.4.0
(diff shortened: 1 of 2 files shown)


none0nfg:
Just override appProtocol value to "", to disable it

none0nfg:
@pierluigilenoci
I've tried to set `appProtocol: ""` in values.yaml, but k8s service gets the same `appProtocol: ""`, and so it's raising a error. 
The way with if, does'nt add appProtocol directive to service notation at all.

pierluigilenoci:
Can you check the manifests generated by the Chart?
What you ask is already a feature of the chart.

pierluigilenoci:
It was closed due to lack of feedback from the reporter.

Issue #142, opened 2023-04-12
Provide good securityContext by default


jemag:
**Describe the solution you'd like**
Provide good default for the various securityContext like:
```yaml
  securityContext:
    allowPrivilegeEscalation: false
    capabilities:
      drop:
      - ALL
    readOnlyRootFilesystem: true
    runAsGroup: 2000
    runAsNonRoot: true
    runAsUser: 2000
    seccompProfile:
      type: RuntimeDefault

```

Currently securityContext is not enabled, although some values are potentially suggested in comment, e.g:
```yaml
securityContext:
  enabled: false
  runAsNonRoot: true
  # allowPrivilegeEscalation: false
  # runAsUser: 2000
```
**What is the added value?**
This would improve the default security out of the box for helm chart users. If the containers currently support it, there isn't much downside to improving the default security.

It also helps clarifying to the end-user that these values are officially supported and will not cause any problem with the containers (e.g.: `readOnlyRootFilesystem: true` could for example cause problems for containers expecting to write to specific directories). By having them already specified, the end-user does not need to do extensive testing to ensure it does not cause any issue.

jemag:
I can submit a PR with improved default securityContext if you would like

pierluigilenoci:
@jemag It would be nice if you create a PR for this.
Even if changing a chart's default configuration is risky, please consider it carefully.

PR #145, opened 2023-04-21
Set strong securityContext by default

jemag:
Fixes https://github.com/oauth2-proxy/manifests/issues/142

I have been using this `securityContext` with 3 different oauth-proxy so far and have not encountered any issues. Would still appreciate someone other than me testing it as well.

### DIFF
--- a/helm/oauth2-proxy/Chart.yaml
+++ b/helm/oauth2-proxy/Chart.yaml
@@ -1,5 +1,5 @@
 name: oauth2-proxy
-version: 6.13.0
+version: 6.14.0
 apiVersion: v2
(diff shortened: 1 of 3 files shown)


pierluigilenoci:
@jemag This is a breaking change that cannot be introduced this way. 
We must avoid breaking previous installations.

jemag:
@pierluigilenoci Do you have any suggestions on how you would like this change to be introduced? 
Should I have incremented the `Major` version instead of `Minor`?

This should not be breaking previous installations and has not in my testing, but let me know if  you think I missed something.

pierluigilenoci (approved):

PR #152, opened 2023-06-07
Allow alpha config to come from a template

### LINKED ISSUE #151: Allow templating of alpha config
It would be good to be able to define some parts of the alpha config as a string that is treated as a Helm template, similar to the normal config.

This is of particular use when using oauth2-proxy as a subchart of a parent chart.

mkjpryor:
Fixes #151

### DIFF
--- a/helm/oauth2-proxy/Chart.yaml
+++ b/helm/oauth2-proxy/Chart.yaml
@@ -1,5 +1,5 @@
 name: oauth2-proxy
-version: 6.12.1
+version: 6.13.0
 apiVersion: v2
 appVersion: 7.4.0
 home: https://oauth2-proxy.github.io/oauth2-proxy/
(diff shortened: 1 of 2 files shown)


pierluigilenoci:
Thank you so much for your PR.

To approve this PR, the Chart version bump and the modification to the README.md to explain how to use the parameter are missing.
Could you please add them?

mkjpryor:
@pierluigilenoci I realised I didn't add the documentation...! Sorry!

For the version bump, are you expecting a minor or patch version bump? This is obviously a backwards compatible change.

mkjpryor:
@pierluigilenoci 

I added the docs and version bump. I assumed a minor version bump, since this is a feature addition and not just a fix.

pierluigilenoci:
That's fine.
Thanks again.

PR #153, opened 2023-06-07
Use a secret instead of a configmap for alpha config

mkjpryor:
This PR addresses the concerns in #104, namely that using the alpha config requires storing the client secret in a configmap, by moving the whole alpha config to a secret.

Once we have a better way of locating secrets, this can be changed.

### DIFF
--- a/helm/oauth2-proxy/Chart.yaml
+++ b/helm/oauth2-proxy/Chart.yaml
@@ -1,5 +1,5 @@
 name: oauth2-proxy
-version: 6.13.1
+version: 6.13.2
 apiVersion: v2
 appVersion: 7.4.0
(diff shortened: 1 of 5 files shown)


mkjpryor:
Note that the patch version was bumped rather than the minor version as:

  1. The change is backwards compatible
  2. I consider exposing secret information to be a bug

pierluigilenoci:
@mkjpryor, could you please rebase?

mkjpryor:
@pierluigilenoci rebase done

pierluigilenoci (approved):

pierluigilenoci:
@jessebot I was waiting for a review from @JoelSpeed.
Ref: https://github.com/oauth2-proxy/manifests/issues/104#issuecomment-1233395397

PR #159, opened 2023-07-12
Improved securityContext and timeout of the initContainer wait-for-redis

pierluigilenoci:
Added securityContext and timeout of the initContainer wait-for-redis configurable via values


helm/oauth2-proxy/Chart.yaml
```diff
@@ -1,5 +1,5 @@
 name: oauth2-proxy
-version: 6.15.2
+version: 6.15.3
```
desaintmartin, line 2: Suggestion:
```suggestion
version: 6.16.0
```

PR #164, opened 2023-10-15
Fix hostAliases to support multiple entries

saravanan30erd:
**Issue**: 

Current `hostAliases` [implementation](https://github.com/oauth2-proxy/manifests/blob/main/helm/oauth2-proxy/templates/deployment.yaml#L56) supports only one alias entry.

It doesn't support to add two or more alias entries in /etc/hosts.

Refer: https://kubernetes.io/docs/tasks/network/customize-hosts-file-for-pods/#adding-additional-entries-with-hostaliases
We need flexibility to add multiple alias entries.

### DIFF
--- a/helm/oauth2-proxy/Chart.yaml
+++ b/helm/oauth2-proxy/Chart.yaml
@@ -1,5 +1,5 @@
 name: oauth2-proxy
-version: 6.18.0
+version: 6.18.1
 apiVersion: v2
 appVersion: 7.5.1
 home: https://oauth2-proxy.github.io/oauth2-proxy/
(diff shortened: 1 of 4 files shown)


pierluigilenoci:
This is a breaking change and cannot be introduced like this. 
Please [bump the chart](https://github.com/oauth2-proxy/manifests/blob/main/helm/oauth2-proxy/Chart.yaml#L2) accordingly, add the [artifact hub note](https://github.com/oauth2-proxy/manifests/blob/main/helm/oauth2-proxy/Chart.yaml#L37-L41) and update the documentation so it's clear to anyone updating how to implement this.

saravanan30erd:
@pierluigilenoci bumped the version and added artifact nodes, also updated readme and added example in default values.yaml file.

pierluigilenoci (approved):

pierluigilenoci (dismissed):

saravanan30erd:
> @saravanan30erd, can you please rebase your PR?

@pierluigilenoci Looks like you already rebased it. If there is anything I need to do?

pierluigilenoci:
@saravanan30erd I'm waiting for a second opinion from @desaintmartin because this is a breaking change and must be taken care of.

pierluigilenoci (dismissed):

pierluigilenoci (approved):

Issue #108, opened 2022-09-08
Redis subchart outdated


lathspell:
The current Redis subchart points to a chart which uses Redis 6.2 although their latest version is 7.0.
(I have no specific reasons to want the 7.x version, just noticed that a couple of CVEs seems to be fixed in Redis meanwhile)

See 
* https://github.com/oauth2-proxy/manifests/blob/main/helm/oauth2-proxy/Chart.yaml
* https://github.com/bitnami/charts/blob/master/bitnami/redis/Chart.yaml
* https://github.com/redis/redis/releases

pierluigilenoci:
@lathspell would you be kind enough to open a PR?

lathspell:
https://github.com/oauth2-proxy/manifests/pull/109

pierluigilenoci:
@lathspell thank you a lot. Your PR needs to be reworked a little.

pierluigilenoci:
Solved by #165

PR #165, opened 2023-10-17
Bump: Redis chart version bumped to 18.11.5

jgmartinez:
As mentioned in #108 the Redis subchart version is outdated and needs a bump.

This solves several security issues as well as other minor stuff. It's been tested with the Proxy and seems to work properly in our environment.


CasperGN:
Thanks for this merge @pierluigilenoci!

PR #170, opened 2023-11-12
helm: Enrich serviceMonitor properties

jkroepke:
Required for custom relabeling actions or istio deployments.

PR #171, opened 2023-11-13
Fixed the case for serviceMonitor values section

pierluigilenoci:


ldunkum:
Nice PR, but shouldn't this be marked as a breaking change? It can be seen in the parallel PR #170 that this needs some changes to code and also in all deployments that change the values.

PR #189, opened 2024-03-05
removed v from version of oauth2-proxy deployment.yaml

RadOctocode:
Reasoning: assuming all versions for the deployment image for oauth2-proxy will start with v is bad practice.


pierluigilenoci:
@RadOctocode Thanks for your contribution, but some work remains to be considered.
Simply removing that character isn't enough. The code in [other parts](https://github.com/oauth2-proxy/manifests/blob/main/helm/oauth2-proxy/templates/_helpers.tpl) of the [chart](https://github.com/search?q=repo%3Aoauth2-proxy%2Fmanifests%20oauth2-proxy.version&type=code) must also be changed.
Furthermore, the chart needs a version bump.

pierluigilenoci:
@RadOctocode Eventually, can you link me to some documents explaining the rationale behind this change and the technical reasons for this best practice?

RadOctocode:
Hi Pierluigi, sorry about the wait.  I'm happy to make the changes to the other files and bumping the version to remove the v from the helm chart. As far as the reasoning for removing the implicit v from the helm chart, many companies host their images in private docker repositories that have rules. As part of a financial institution one of those rules for our private registry is that image semver flags are not prefixed with a v.

RadOctocode:
Hey @pierluigilenoci I removed the stripping of v from the helper function as well, please let me know if there is anything else I should do.

pierluigilenoci:
@RadOctocode, the chart needs a version bump.

RadOctocode:
@pierluigilenoci I bumped the chart up by one minor version, let me know if I should change this or if I should add something to artifacthub.io/changes

pierluigilenoci:
@RadOctocode, could you please also add the annotation for ArtifactHub?
https://github.com/oauth2-proxy/manifests/blob/main/helm/oauth2-proxy/Chart.yaml#L36-L41

RadOctocode:
@pierluigilenoci I have edited the annotation for artifacthub

pierluigilenoci (approved):

PR #193, opened 2024-04-08
Docs fix for existingSecret needed fields

gcaracuel:
Link to `secret.yaml` will not actually clarify which fields will you need to provide an existing secret but the helper/function rendering secret data.


pierluigilenoci:
@gcaracuel, a chart bump is needed and `artifacthub.io/changes` need to be changed.

pierluigilenoci (approved):

PR #141, opened 2023-04-07
Add proxyVarsFromSecret: "" option

jan-brychta:
added option to load OAUTH2_PROXY variables directly from a secret

**Reasoning**
ExternalSecrets with GitLab provider makes this chart really hard to use, since you cannot name a GitLab CI/CD variable using "-" such as `cookie-secret`.

PR #196, opened 2024-04-09
feat: add proxyVarsFromSecret value

aslafy-z:
Replaces #141


pierluigilenoci:
@tuunit What do you think of this proposal?

helm/oauth2-proxy/Chart.yaml
```diff
@@ -1,5 +1,5 @@
 name: oauth2-proxy
-version: 7.10.4
+version: 7.10.5
```
tuunit, line 2: I would rate this as a new feature and not a patch. So it should be:

```suggestion
version: 7.11.0
```

helm/oauth2-proxy/templates/deployment.yaml
```diff
@@ -172,6 +172,11 @@ spec:
         lifecycle:
 {{ toYaml .Values.lifecycle | indent 10 }}
 {{- end }}
+        {{- with .Values.proxyVarsFromSecret }}
+        envFrom:
+        - secretRef:
+            name: {{  . }}
+        {{- end }}
         env:
         {{- if .Values.proxyVarsAsSecrets }}
```
tuunit, line 181: proxyVarsAsSecrets and proxyVarsFromSecret should be mutually exclusive. Therefore I would like to see this check extend that proxyVarsAsSecrets aren't rendered with proxyVarsFromSecret is set

pierluigilenoci: This PR was given to us as a "gift"; if we want to bring it in, we must introduce the changes we want. 🤷🏻‍♂️

helm/oauth2-proxy/templates/deployment.yaml
```diff
@@ -237,7 +242,7 @@ spec:
         {{- if .Values.extraEnv }}
 {{ tpl (toYaml .Values.extraEnv) . | indent 8 }}
         {{- end }}
-        {{- if .Values.envFrom }}
+        {{- if and (not .Values.envFromExistingSecret) .Values.envFrom }}
         envFrom:
```
tuunit, line 246: @pierluigilenoci I refactored the PR a little bit and gave the variable a better name but then I realised we already have an `envFrom` parameter that allows for using a secret as a source like so:

```
envFrom:
  - secretRef:
      name: my-secret
```

in the `values.yaml` without any changes to the chart.

tuunit: Therefore I don't see the benefit of this addition to the helm chart 🤔

pierluigilenoci: The original idea of PR has probably been lost over time. 
However, you can find it in this comment:
https://github.com/oauth2-proxy/manifests/pull/141#issuecomment-1514677797

tuunit: Since then it has been replaced with a more generic solution:
https://github.com/oauth2-proxy/manifests/pull/201

tuunit: I would therefore close this PR.

Issue #182, opened 2024-01-09
Wait-for-redis cannot handle redis in deployment mode


Pionerd:
The `wait-for-redis` container checks for the redis-master container to be ready, but the name of the container is semi-hardcoded as `pod/{{ include "oauth2-proxy.redis.fullname" . }}-master-0` (https://github.com/oauth2-proxy/manifests/blob/0bdde21382538e65b6c9896858fdbf7f986464d4/helm/oauth2-proxy/templates/deployment.yaml#L69C34-L69C34). This is the format of a StatefulSet. 

We would like to use redis for the sole purpose of saving the cookies, but don't mind if they get lost when redis would be restarted or killed, so we prefer to run stateless without PV, so we could run redis-master in deployment mode. This results in a pod named `oauth2-proxy-redis-master-{uuid}`, so the current check does not pass. 

As I understand, it's not possible to use a kubectl wait on a service, so we will need to determine the name of the pod to check differently.

Before writing a pull request, what do you guys think of `kubectl -n oauth2-proxy wait --for=condition=ready --selector=app.kubernetes.io/name=redis --selector=app.kubernetes.io/component=master po`? Where we could make the selectors configurable?

pierluigilenoci:
I would say that being able to control the readiness status of the pod via selectors is certainly an evolution of the current solution, as it could lead to unwanted results if someone abuses the labels used. 🤷🏻‍♂️ 

So, I would say it all depends on how the solution is implemented. 👍🏻

Pionerd:
It depends a bit a bit on how far we want to go. In my case (using the reds sub chart) the selector for the service is: 
```
app.kubernetes.io/component: master        # Hardcoded in redis subchart                                                                                                                                                                                         
app.kubernetes.io/instance: oauth2-proxy  # {{ .Release.Name }}                                                                                                                                                                                           app.kubernetes.io/name: redis                      # subchart name 
```
I think we could generate the above as the default selectors with the possibility of overwriting this via `values.yaml`.

mvalenzisi:
I agree with the solution proposed, but how can we handle the case when Redis is installed with Sentinel?
In that case, the pods have the name `oauth2-proxy-redis-node-<n>` and labels:
```
app.kubernetes.io/component=node
app.kubernetes.io/instance=oauth2-proxy
app.kubernetes.io/name=redis
```

Pionerd:
[...]

Issue #191, opened 2024-03-22
Redis sentinel does not work, chart throws error on install


sarg3nt:
The helm chart throws an error during install with `sessionStorage.rdis.clientType: sentinel`
I've tried all kinds of configs and cannot get it to run properly.  
I haven't been able to determine if there is a value I'm not setting that it doesn't like or if it's just broken.

**Error:**
`Deployment in version "v1" cannot be handled as a Deployment: json: cannot unmarshal array into Go struct field EnvVar.spec.template.spec.containers.env.value of type strin`

**Config:**
```yaml
config:
  clientID: <redacted>
  clientSecret: <redacted>
  configFile: |-
    pass_basic_auth = false
    pass_access_token = true
    set_authorization_header = true
    pass_authorization_header = true
    reverse_proxy = true
    whitelist_domains = [".bob.com"]
    email_domains = [
      "*",
    ]
    allowed_groups = [
      "redacted"
    ]
    azure_tenant = "redacted"
    oidc_issuer_url = "https://redacted"
    provider = "azure"
    upstreams = ["file:///dev/null"]
    http_address = "0.0.0.0:4180"
    cookie_domains = ["k8s.bob.com"]
  cookieSecret: redacted
extraArgs:
  redis-insecure-skip-tls-verify: "true"
extraEnv:
- name: HTTP_PROXY
   value: <redacted along with rest of env vars>
image:
  pullPolicy: Always
ingress:
  annotations:
    ingress.kubernetes.io/ssl-redirect: "true"
    kubernetes.io/ingress.class: nginx
    kubernetes.io/tls-acme: "true"
  enabled: true
  hosts:
  - oauth2.bob.com
  path: /
  tls:
  - hosts:
    - oauth2.bob.com
    secretName: oauth2-proxy-tls
redis:
  enabled: true
  global:
    imageRegistry: <redacted>
    redis:
      password: <redacted>
  sentinel:
    enabled: true
  tls:
    authClients: false
    autoGenerated: true
    enabled: true
replicaCount: 3
resources:
  limits:
    cpu: 500m
    memory: 300Mi
  requests:
    cpu: 10m
    memory: 30Mi
sessionStorage:
  redis:
    clientType: sentinel
    password: <redacted>
    sentinel:
      # The following are fake, I've also tried sending it a TF array.
      connectionUrls: '["redis://127.0.0.1:8000","redis://127.0.0.1:8000","redis://127.0.0.1:8000"]'
      masterName: master
[...]

PR #197, opened 2024-04-09
Improve wait-for-redis logic

Pionerd:
Fixes: https://github.com/oauth2-proxy/manifests/issues/182

Tested the standalone mode OK (both with Deployments and Statefulsets, which was my original use case)

Cannot test sentinel and cluster, because I also run into https://github.com/oauth2-proxy/manifests/issues/191


pierluigilenoci (commented):
Can you please also edit the Chart.yaml file bumping the chart version and editing the `artifacthub.io/changes` section?

pierluigilenoci:
To test the cases covered by #191 you can try using `[]`

Pionerd:
Done

pierluigilenoci (approved):

PR #208, opened 2024-05-27
add autoscaling

sja:


pierluigilenoci:
@sja, please add a helm chart bump and edit the artifacthub.io/changes section.

sja:
Thanks @pierluigilenoci  for that info. I did a minor version bump, as it adds a feature. It is backwards compatible, old values do not change the result.

sja:
@pierluigilenoci I updated the MR again, inspired from the other MRs. Feel free to reach out to me, especially if you want me to modify the version numbers to not have a conflict. Usually, I react in 1 business day.

sja:
I added that, but I'm not familar with how to validate the result. Are the results compared to an expected value somewhere or is that more like a smoke test if the values can be rendered at all?

pierluigilenoci (approved):

pierluigilenoci:
@sja Usually, the test results are visible in the [pipeline](https://github.com/oauth2-proxy/manifests/actions/runs/9301491337/job/25609773690). 
In this case, they are not, and I don't understand why.
However, from the point of view of the feature, this PR is OK like this.

pierluigilenoci:
@sja ok, fixed https://github.com/oauth2-proxy/manifests/pull/211 🚀

pierluigilenoci:
@sja Thank you.

PR #210, opened 2024-05-30
fix: use the correct default configmap name in deployment

BitForger:
If you wish to use the default config map and only supply emails in `authenticatedEmailsFile.restricted_users` option, the chart doesn't include the proper reference to the default config map name. See below:

https://github.com/oauth2-proxy/manifests/blob/6116131b136697e10dda185eaed2bdda347b2d33/helm/oauth2-proxy/templates/configmap-authenticated-emails-file.yaml#L13

### DIFF
--- a/helm/oauth2-proxy/Chart.yaml
+++ b/helm/oauth2-proxy/Chart.yaml
@@ -1,5 +1,5 @@
 name: oauth2-proxy
-version: 7.6.1
+version: 7.6.3
 apiVersion: v2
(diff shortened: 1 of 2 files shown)


pierluigilenoci:
@BitForger, please add a helm chart bump and edit the artifacthub.io/changes section.

pierluigilenoci (approved):

pierluigilenoci:
@BitForger Thank you.

PR #216, opened 2024-06-14
fix: path in volume fixed

dklein-net:
The path in the volume "configaccesslist" was wrong. I've changed this corresponding to the parameter in the startup args.

### DIFF
--- a/helm/oauth2-proxy/Chart.yaml
+++ b/helm/oauth2-proxy/Chart.yaml
@@ -1,5 +1,5 @@
 name: oauth2-proxy
-version: 7.7.3
+version: 7.7.4
 apiVersion: v2
 appVersion: 7.6.0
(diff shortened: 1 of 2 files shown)


pierluigilenoci:
@dklein-net, please add a [chart version bump](https://github.com/oauth2-proxy/manifests/blob/main/helm/oauth2-proxy/Chart.yaml#L2) and adequately change the [artifact hub changelog message](https://github.com/oauth2-proxy/manifests/blob/main/helm/oauth2-proxy/Chart.yaml#L36-L41).

dklein-net:
@pierluigilenoci thanks for the hint. I've added a version bump and changed the artifact hub annotation.

pierluigilenoci (approved):

pierluigilenoci:
@dklein-net Thank you.

PR #217, opened 2024-06-24
fix: Deployment name in HPAs scaleTargeRef spec

losil:
The Deployment name in `helm/oauth2-proxy/templates/deployment.yaml` is configured with `{{ template "oauth2-proxy.fullname" . }}` which can cause `Deployment not found` issues when the default Release name is overwritten.


pierluigilenoci (approved):

pierluigilenoci:
@losil Thank you.

PR #220, opened 2024-06-27
fix: Reorder metrics list in HPA

losil:
This fixes reconciliation loops in Argo CD because HPA controller reorders the list entries.
The observed behaviour is described in the following issue: https://github.com/kubernetes/kubernetes/issues/74099


pierluigilenoci (approved):

pierluigilenoci:
@losil Thank you.

PR #255, opened 2024-11-20
chore: Add missing executable permission for check-redis.sh

T-vK:
I added the missing executable permission by running `chmod +x ./check-redis.sh`
```
Before: -rw-r--r-- / 644
After:  -rwxr-xr-x / 755
```


pierluigilenoci:
@T-vK, please add a chart version bump with a proper ArtifacHub message.

helm/oauth2-proxy/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 name: oauth2-proxy
-version: 7.7.30
+version: 7.7.31
 apiVersion: v2
-appVersion: 7.7.1
+appVersion: 7.7.2
```
tuunit, line 4: Please revert. There is no v7.7.2 for oauth2-proxy as of yet

tuunit (changes requested):

helm/oauth2-proxy/Chart.yaml
T-vK: Done.

tuunit (approved):

PR #256, opened 2024-11-20
feat: allow templates in annotations

### LINKED ISSUE #254: Allow using templates in annotations
I have a use case where I need to include helm templates in an annotation. Example:

```yaml
  ingress: 
    enabled: true
    annotations:
      nginx.ingress.kubernetes.io/custom-http-errors: "401,403,404,500,502,503"
      nginx.ingress.kubernetes.io/default-backend: "{{ .Release.Name }}-error-pages"
```

This does not work as of now - the "{{ .Release.Name }}"-part is not interpreted and written as is to the rendered manifest.

gtudan:
This allows templates in annotations for all resources. It tried to keep the pattern used in other places (i.e. extra-env for deployment) and use the same pattern for all annotations.

/fixes #254


helm/oauth2-proxy/ci/tpl-values.yaml
```diff
@@ -8,6 +8,8 @@ ingress:
   tls:
     - hosts:
         - "{{ $.Release.Name }}.local"
+  annotations:
+    nginx.ingress.kubernetes.io/default-backend: "{{ $.Release.Name }}-default-backend"
```
pierluigilenoci, line 12: Why this? You don't need to add this.

helm/oauth2-proxy/templates/secret-alpha.yaml
```diff
@@ -7,8 +7,8 @@
 apiVersion: v1
 kind: Secret
 metadata:
-{{- if .Values.alphaConfig.annotations }}
-  annotations: {{- toYaml .Values.alphaConfig.annotations | nindent 4 }}
+{{- with .Values.alphaConfig.annotations }}
+{{ tpl ( toYaml . ) $ | indent 4 }}
```
pierluigilenoci, line 11: Can you put it back?
```
annotations:
```

pierluigilenoci (changes requested):
There are a few things to fix. 
Also missing are the chart version bump and release notes for ArtifactHub.

pierluigilenoci:
I approved the tests; at least you can see which tests failed.

helm/oauth2-proxy/templates/secret-alpha.yaml
gtudan: Sorry, copy-pasted to much - it's back, thanks for noticing!

helm/oauth2-proxy/ci/tpl-values.yaml
gtudan: I have to confess I don't fully understand how the tests work - I added this so I could see if the template is being resolved and to ensure the linter won't complain about the format.

If you can give me a hint on where to add a separate test case for this, then I'm happy to do so :-)

gtudan:
>  Also missing are the chart version bump and release notes for ArtifactHub.

I incremented the chart version and added the changes as annotations - hope I got the format right?

Sorry, I checked for a contributing instruction, but didn't find one...

gtudan:
@pierluigilenoci  The last merge introduced an extra newline and the linter complained about it. I removed it and pushed again. Could you please approve the workflow again?

tuunit:
Approved the run

helm/oauth2-proxy/ci/tpl-values.yaml
pierluigilenoci: @gtudan to test the chart we use ct.
Specifically this https://github.com/helm/chart-testing/blob/main/doc/ct_lint.md
> Charts may have multiple custom values files matching the glob pattern '*-values.yaml' in a directory named 'ci' in the root of the chart's directory. The chart is linted for each of these files. If no custom values file is present, the chart is linted with defaults.

pierluigilenoci (approved):

helm/oauth2-proxy/ci/tpl-values.yaml
pierluigilenoci: @gtudan, I have already approved the PR, but if you want to add to the tests all the cases you added in the PR, that would be great, so we have them all covered.

pierluigilenoci: Ok, I did it for you. 😄

pierluigilenoci:
> I incremented the chart version and added the changes as annotations - hope I got the format right?

You did well 😄 

> Sorry, I checked for a contributing instruction, but didn't find one...

In reality, a contribution guide is missing.

tuunit (approved):

PR #268, opened 2024-11-27
fix: connection urls to be automatically comma separated

tuunit:
## Description
The `connectionUrls` attribute of `sessionStorage.redis.cluster` and `sessionStorage.redis.sentinel` is currently of type array which doesn't work.

## Solution
Add `join ","` to the templating to automatically generate a comma separated string and allow for single string values as well.

### DIFF
--- a/helm/oauth2-proxy/ci/redis-sentinel-comma-values.yaml
+++ b/helm/oauth2-proxy/ci/redis-sentinel-comma-values.yaml
@@ -1,0 +1,22 @@
+sessionStorage:
+  type: redis
+  redis:
+    clientType: sentinel
+    sentinel:
+      password: "foo"
+      masterName: "mymaster"
+      connectionUrls: "redis://oauth2-proxy-redis:26379,redis://oauth2-proxy-redis:26379"
+redis:
+  # provision an instance of the redis sub-chart
+  enabled: true
+  fullnameOverride: oauth2-proxy-redis
+  architecture: replication
+  sentinel:
+    enabled: true
+    masterSet: mymaster
+  global:
+    redis:
+      password: "foo"
+initContainers:
+  waitForRedis:
+    enabled: true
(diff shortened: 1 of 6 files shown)


tuunit:
@pierluigilenoci I think the new test case is broken. Will quickly check tomorrow. I have a bad connection right now and cannot fetch the redis helm chart -.-

tuunit:
> @tuunit, this is a breaking change. The version bump must be a major version and a migration guide must be added.

This is not a breaking change!!

pierluigilenoci:
@tuunit, do you have any chances to move forward with this?

pierluigilenoci (dismissed):

pierluigilenoci (dismissed):

tuunit:
@pierluigilenoci please have another review of this PR :)

I added two test cases to demonstrate that it is not a breaking change

helm/oauth2-proxy/ci/redis-sentinel-comma-values.yaml
```diff
@@ -0,0 +1,22 @@
+sessionStorage:
+  type: redis
+  redis:
+    clientType: sentinel
+    sentinel:
+      password: "foo"
+      masterName: "mymaster"
+      connectionUrls: "redis://oauth2-proxy-redis:26379,redis://oauth2-proxy-redis:26379"
```
pierluigilenoci, line 8: The previous example was ```["redis://127.0.0.1:8000", "redis://127.0.0.1:8000"]``` so it was not a string.

tuunit: Yes I know but that is exactly the point of this PR that was never possible:

Following test are done on the current main branch:

image

The example mentioned the following syntax:

```yaml
(e.g. `["redis://127.0.0.1:8000", "redis://127.0.0.1:8000"]`)
```

So like this:

image

And even in single quotes doesn't work:
image

PR #276, opened 2025-01-14
chore: update oauth2-proxy to v7.8.0

air3ijai:
New version of oauth2-proxy was just released with CVE fixes and other changes.

https://github.com/oauth2-proxy/oauth2-proxy/releases/tag/v7.8.0

### DIFF
--- a/helm/oauth2-proxy/Chart.yaml
+++ b/helm/oauth2-proxy/Chart.yaml
@@ -1,7 +1,7 @@
 name: oauth2-proxy
-version: 7.9.2
+version: 7.10.0
 apiVersion: v2
-appVersion: 7.7.1
+appVersion: 7.8.0
 home: https://oauth2-proxy.github.io/oauth2-proxy/
 description: A reverse proxy that provides authentication with Google, Github or other providers
 keywords:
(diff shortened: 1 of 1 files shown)


tuunit:
@air3ijai please update the artifact hub changelog message as well. in line 36 and below :)

Thanks for the quick action.

helm/oauth2-proxy/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 name: oauth2-proxy
-version: 7.9.2
+version: 7.9.3
```
tuunit, line 2:
```suggestion
version: 7.10.0
```

it's not a patch really should be a minor bump instead

tuunit (changes requested):

helm/oauth2-proxy/Chart.yaml
```diff
@@ -35,7 +35,7 @@ kubeVersion: ">=1.16.0-0"
 annotations:
   artifacthub.io/changes: |
     - kind: fixed
```
tuunit, line 37:
```suggestion
    - kind: changed
```

PR #282, opened 2025-02-05
doc(maintainers): update maintainers in Chart.yaml

tuunit:
@pierluigilenoci I don't think we should increase the version for this change, right?

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -6,6 +6,7 @@
   automerge: false,
   assignees: [
     'pierluigilenoci',
+    'tuunit',
   ],
   schedule: [
     'before 5am',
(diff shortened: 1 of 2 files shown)


tuunit:
> I think that, unfortunately, it is a good change. I'll still propose new PRs here and there ! 

Do you want to stay in the list? I would be open to it :)

pierluigilenoci:
@tuunit why do you change maintainers? 

The chart version bump is necessary in the pipeline whenever there is a change in any of the chart files. 
So, I assume the version bump is needed.

pierluigilenoci (dismissed):

pierluigilenoci (approved):

tuunit:
> @tuunit why do you change maintainers?
> 
> The chart version bump is necessary in the pipeline whenever there is a change in any of the chart files. So, I assume the version bump is needed.

@pierluigilenoci 
Because Joel and removed all members of the organization  that haven't been involved in the project in the past 1 - 1.5 years.

If it is okay with you I would remove your last commit and just overrule the pipeline for this merge and don't trigger a release for this change.

pierluigilenoci:
@tuunit, please refer to my comment here: https://github.com/oauth2-proxy/manifests/pull/283#issuecomment-2639483918

PR #283, opened 2025-02-05
doc: improved the documentation in values.yaml for Redis password

max-pfeiffer:
I made the password usage for Redis sub chart more clear that way.

### DIFF
--- a/helm/oauth2-proxy/Chart.yaml
+++ b/helm/oauth2-proxy/Chart.yaml
@@ -1,5 +1,5 @@
 name: oauth2-proxy
-version: 7.10.4
+version: 7.10.5
 apiVersion: v2
 appVersion: 7.8.1
 home: https://oauth2-proxy.github.io/oauth2-proxy/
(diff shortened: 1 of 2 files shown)


pierluigilenoci:
Even if they are just comments, a version bump for the chart is needed. 😓

tuunit:
> Even if they are just comments, a version bump for the chart is needed. 😓

@pierluigilenoci 
Maybe we should consider changing the pipeline a bit for such changes or I can just overrule the pipeline if necessary?

pierluigilenoci:
@tuunit In my experience, there is no way in the pipeline to understand whether the change introduced in the chart files is just a comment.
You can do as you like, but as far as I'm concerned, it's OK to bump even when we change just one comment. In the end, the chart is different from the previous one, so a "patch" increase makes sense.

tuunit:
> @tuunit In my experience, there is no way in the pipeline to understand whether the change introduced in the chart files is just a comment. You can do as you like, but as far as I'm concerned, it's OK to bump even when we change just one comment. In the end, the chart is different from the previous one, so a "patch" increase makes sense.

It is possible we do have the releases and tags and could compare the content of the current Chart.yaml against the latest tag and check if the version has changed and only if so run the actual release process. Might not be necessary for such a small chart.

pierluigilenoci (approved):

pierluigilenoci (approved):

tuunit (approved):

PR #301, opened 2025-04-27
Solve the problem with upstreams being prohibited when alphaconfig is enabled

### LINKED ISSUE #226: failed to load alpha options: unable to load config file: read /etc/oauth2_proxy/oauth2_proxy.yml: is a directory
In AKS, Pod fails to start with the error mention in the tittle:

```shell
[main.go:41] ERROR: failed to load alpha options: unable to load config file: read /etc/oauth2_proxy/oauth2_proxy.yml: is a directory
```

It is installed via terraform, with the following values overrides:

values file:

```yaml
config:
  configFile: |-
    email_domains = [ "*" ]        # Restrict to these E-Mail Domains, a wildcard "*" allows any email

extraVolumes: ${jsonencode(extra_volumes)} # CSI driver volume
extraVolumeMounts: ${jsonencode(extra_volume_mounts)} #Mounts to /mnt/secret
alphaConfig:
  enabled: true
  existingSecret: ${oauth2_secret}
  configData:
    providers:
    - id: oicd-azure
      provider: oidc
      azureConfig:
        tenant: ${tenant_id}
      oidcConfig:
        issuerURL: https://login.microsoftonline.com/${tenant_id}/v2.0
        jwksURL: https://login.microsoftonline.com/common/discovery/v2.0/keys
        userIDClaim: oid
        audienceClaims: [aud]
        emailClaim: email
        groupsClaim: groups
    upstreamConfig:
      upstreams:
        - id: static_200
          path: /
          static: true
          staticCode: 200
    injectResponseHeaders:
      - name: X-Auth-Request-Preferred-Username
        values:
          - claim: preferred_username
      - name: X-Auth-Request-Email
        values:
          - claim: email
      - name: X-Auth-Request-Id-Token
        values:
          - claim: id_token
      - name: X-Auth-Request-Groups
        values:
          - claim: groups

extraArgs:
  reverse-proxy: true
  skip-provider-button: true 
  silence-ping-logging: true
  cookie-refresh: "15m"
  cookie-expire: "24h"

redis:
  enabled: false

sessionStorage:
  type: redis
  redis:
    existingSecret: redis-settings
    standalone:
        connectionUrl:  "<redacted>"
```

This started to happen after I upgraded from 6.23.1 to the more recent 7.6.0
I have also ensured it is using the latest chart version (7.7.9) and verified the structure of values.yaml to match with the latest chart version.

If I omit configFile from config section, I get:

``` shell
 failed to load core options: failed to load config: error unmarshalling config: 1 error(s) decoding:
* '' has invalid keys: upstreams
```
So, `config.configFile.upstreams = [ "file:///dev/null" ]` seems to be invalid. It breaks when configFile is not overriden.

I am still unable to upgrade oauth2-proxy to use latest chart and image versions. But still investigating if I can workaround the issue. I suspect this has something to do with how newer versions treat multiple provider configurations that may not be reflected in the chart, even though I am only using a single provider in alphaConfiguration.

EvanCarroll:
Upstreams is no longer allowed if you have alphaConfig enabled, because the option is available in alphaConfig. However, setting the default in alphaConfig won't work, because people may have it disabled, and it's disabled by default.

This means we need this value to remain in configFile as a default if alphaConfig isn't enabled.

https://github.com/oauth2-proxy/manifests/issues/226

### DIFF
--- a/helm/oauth2-proxy/templates/configmap.yaml
+++ b/helm/oauth2-proxy/templates/configmap.yaml
@@ -13,6 +13,13 @@ metadata:
   name: {{ template "oauth2-proxy.fullname" . }}
   namespace: {{ template "oauth2-proxy.namespace" $ }}
 data:
+  {{- if and (not .Values.alphaConfig.enabled) .Values.config.legacyDefaults }}
+  oauth2_proxy.cfg: |-
+{{ tpl .Values.config.legacyDefaults $ | indent 4 }}
+
+{{ tpl .Values.config.configFile $ | indent 4 }}
+  {{- else }}
   oauth2_proxy.cfg: {{ tpl .Values.config.configFile $ | quote }}
+  {{- end }}
 {{- end }}
 {{- end }}
--- a/helm/oauth2-proxy/values.yaml
+++ b/helm/oauth2-proxy/values.yaml
@@ -43,10 +43,18 @@ config:
     # Example:
     #  - group1@example.com
     #  - group2@example.com
+
+  # these exist only because if alphaConfig.enabled is not set then the
+  # defaults there need to be present here. If it is set, then the defaults can
+  # not be here because they're not valid. This should go away when alphaConfig
+  # is default on.
+  # https://github.com/oauth2-proxy/manifests/issues/226
+  legacyDefaults: |-
+    upstreams = [ "file:///dev/null" ]
+
   # Default configuration, to be overridden
   configFile: |-
     email_domains = [ "*" ]
-    upstreams = [ "file:///dev/null" ]
   # Custom configuration file: oauth2_proxy.cfg
   # configFile: |-
   #   pass_basic_auth = false
@@ -64,7 +72,9 @@ alphaConfig:
   # Arbitrary configuration data to append to the metrics section
   metricsConfigData: {}
   # Arbitrary configuration data to append
-  configData: {}
+  configData:
+    upstreams:
+      - "file:///dev/null"
   # Arbitrary configuration to append
   # This is treated as a Go template and rendered with the root context
   configFile: ""
--- a/helm/oauth2-proxy/Chart.yaml
+++ b/helm/oauth2-proxy/Chart.yaml
@@ -1,5 +1,5 @@
 name: oauth2-proxy
-version: 7.12.10
+version: 7.12.11
 apiVersion: v2
 appVersion: 7.8.2
 home: https://oauth2-proxy.github.io/oauth2-proxy/
@@ -32,7 +32,7 @@ kubeVersion: ">=1.16.0-0"
 annotations:
   artifacthub.io/changes: |
     - kind: changed
-      description: Updated the Redis chart to the latest version
+      description: Solve the problem with upstreams being prohibited when alphaconfig is enabled
       links:
         - name: Github PR
-          url: https://github.com/oauth2-proxy/manifests/pull/303
+          url: https://github.com/oauth2-proxy/manifests/pull/301


helm/oauth2-proxy/values.yaml
```diff
@@ -64,7 +72,9 @@ alphaConfig:
   # Arbitrary configuration data to append to the metrics section
   metricsConfigData: {}
   # Arbitrary configuration data to append
-  configData: {}
+  configData:
+    upstreams:
+      - "file:///dev/null"
```
tuunit, line 77: Why do we have to have it twice? Here and in the `legacyDefaults` block?

helm/oauth2-proxy/values.yaml
```diff
@@ -32,6 +32,15 @@ config:
   # If left empty, it will default to the release name
   cookieName: ""
   google: {}
+
```
tuunit, line 35: Please move the example config for google back here under the correct block. It is currently below your new `legacyDefaults` block

tuunit (changes requested):

helm/oauth2-proxy/values.yaml
EvanCarroll: Because here it's a line of yaml, in legacyDefaults it's a line of toml. The legacy file must be TOML and the alphaConfig is yaml. Without introducing a macro that generate TOML from YAML we have to have it twice.

helm/oauth2-proxy/values.yaml
EvanCarroll: Just fixed it. Please reopen if I missed something.

pierluigilenoci:
@EvanCarroll The PR also lacks the chart bump and the message for ArtifactHub

EvanCarroll:
@pierluigilenoci Would have gladly done it, I wasn't sure if you wanted to release it immediately or not.

EvanCarroll:
@pierluigilenoci that good?

helm/oauth2-proxy/templates/configmap.yaml
```diff
@@ -13,6 +13,13 @@ metadata:
   name: {{ template "oauth2-proxy.fullname" . }}
   namespace: {{ template "oauth2-proxy.namespace" $ }}
 data:
+  {{- if and (not .Values.alphaConfig.enabled) .Values.config.legacyDefaults }}
```
tuunit, line 16: Isn't legacyDefaults always defined through the default values in `values.yaml` and even if a user reconfigures it to be an empty object or empty string wouldn't the templating still work?

I'm just asking myself if we really need the if else condition here? Furthermore if required, I would rather prefer to have it surround the actual templating line for the legacyDefaults instead of the whole `oauth2_proxy.cfg` block

`{{ tpl .Values.config.legacyDefaults $ | indent 4 }}`

tuunit (changes requested):
One more nitpick in terms of code clarity and simplicity

helm/oauth2-proxy/templates/configmap.yaml
EvanCarroll: That would probably work too, but I'm done w/ this one so someone else will have to pick it up.  If I was going to spend any more time on this I would just put the remaining option into alphaConfig so the whole legacyDefaults can go away. This isn't supposed to be a long lasting fix. Just a patch to get you to the next point.

helm/oauth2-proxy/templates/configmap.yaml
```diff
@@ -13,6 +13,13 @@ metadata:
...
+  oauth2_proxy.cfg: |-
+{{ tpl .Values.config.legacyDefaults $ | indent 4 }}
+
+{{ tpl .Values.config.configFile $ | indent 4 }}
```
tuunit, line 20: This will actually cause a breaking change for existing deployments:

```
[2025/04/30 18:30:31] [main.go:42] ERROR: failed to load config: unable to load config file: While parsing config: toml: key upstreams is already defined
```

tuunit (changes requested):

helm/oauth2-proxy/templates/configmap.yaml
tuunit: I appreciate the attempt to help, but after checking this change introduces a breaking change, and we can’t merge it as-is. I know it seems trivial but at the scale oauth2-proxy is being utilized (we are talk 30+ M downloads a month), even minor tweaks need thorough validation! Throwing in a patch and declaring you’re ‘done’ isn’t how stable software gets maintained.

PR #317, opened 2025-06-11
Allow maxUnavaiable in PodDisruptionBudget

pinkavaj:


pinkavaj:
>     1. [They cannot both be specified](https://kubernetes.io/docs/tasks/run-application/configure-pdb/#specifying-a-poddisruptionbudget), so I would prefer that a logic be introduced in the chart that gives priority to one of the two so that if both are selected, only the one with the highest priority is inserted into the PDB. I would say that it needs to be `minAvailable` for backward compatibility.

They cannot be specified both, but setting value to `null` effectively removes the key (tested) and acts like the key is not present, tested and it works. I can add the logic, but I would prefer to fail loudly if the configuration is invalid (which will fail anyway, just a bit latter) instead of letting one of the values silently disappear when both are present.

> 
>     2. The chart version needs to be changed, and the annotation for the ArtifactHub needs to be corrected.

ArtifactHub annotation updated, thx. I have bumped the chart version, but I might have missed something pleas can you be more specific what needs to be changed?

helm/oauth2-proxy/templates/poddisruptionbudget.yaml
```diff
@@ -11,5 +11,6 @@ spec:
   selector:
     matchLabels:
       {{- include "oauth2-proxy.selectorLabels" . | indent 6 }}
+  maxUnavailable: {{ .Values.podDisruptionBudget.maxUnavailable }}
```
tuunit, line 14: As only max unavailable or min available can be set and they are mutually exclusive shouldn't we do something like

```
{{- if .Values.podDisruptionBudget.maxUnavailable }}
maxUnavailable: {{ .Values.podDisruptionBudget.maxUnavailable }}
{{- else }}
minAvailable: {{ .Values.podDisruptionBudget.minAvailable }}
{{- end }}
```

> You can specify only one of maxUnavailable and minAvailable in a single PodDisruptionBudget. maxUnavailable can only be used to control the eviction of pods that have an associated controller managing them.

https://kubernetes.io/docs/tasks/run-application/configure-pdb/

pinkavaj: You are completely right only one of those can be specified, the "trick" is (and it is actually not trick but documented propery) the `null` value in a key is the indicator of "not beeing specified". 

You can try it your-selve by  running `kubectl apply` on the following manifest int K8s cluster:

```
apiVersion: policy/v1
kind: PodDisruptionBudget
metadata:
  name: test
spec:
  maxUnavailable: 30%
  minAvailable: null
  selector:
    matchLabels:
      app.kubernetes.io/component: test
```

I have updated the documentation to make it more obvious and clear to everyone.

I'm agaist the `if ..`, because then the PDB configuration will behave differently compared to K8s, one of the parameters will be magically discarded instead of throwing an error about invalid configuration (the person then can live in expectations both configuration keys were applied, if his knowledge of internal workings are not deep enough)

tuunit: I'm fine with the documentation adaption

PR #321, opened 2025-07-17
chore: update oauth2-proxy to v7.10.0

pierluigilenoci:


helm/oauth2-proxy/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 name: oauth2-proxy
-version: 7.13.1
+version: 7.13.2
```
tuunit:
```suggestion
version: 7.14.1
```

pierluigilenoci:
@tuunit, if you want a minor bump, the correct one is `7.14.0`

tuunit:
> @tuunit, if you want a minor bump, the correct one is `7.14.0`

I wanted to merge the other PR first and already prepared this PR accordingly and then got distracted. Thanks for fixing the conflict!

PR #322, opened 2025-07-21
fix(PDB): Don't set null inside PDB

mkilchhofer:
This PR here:
- https://github.com/oauth2-proxy/manifests/pull/317

introduced a regression for everyone using this Kyverno policy: https://kyverno.io/policies/other/pdb-maxunavailable/pdb-maxunavailable/

```log
validation error: The value of maxUnavailable must be greater than zero. rule deny-pdb-maxunavailable failed at path /spec/maxUnavailable/
```

Since both fields are mutual exclusive, I suggest that we only render the related field according to values:
```bash
$ kubectl explain pdb.spec
(..)
FIELDS:
  maxUnavailable	<IntOrString>
    An eviction is allowed if at most "maxUnavailable" pods selected by
    "selector" are unavailable after the eviction, i.e. even in absence of the
    evicted pod. For example, one can prevent all voluntary evictions by
    specifying 0. This is a mutually exclusive setting with "minAvailable".

  minAvailable	<IntOrString>
    An eviction is allowed if at least "minAvailable" pods selected by
    "selector" will still be available after the eviction, i.e. even in the
    absence of the evicted pod.  So for example you can prevent all voluntary
    evictions by specifying "100%".
(..)
```


tuunit:
@pinkavaj not what I was expecting but my concern was valid.

tuunit (approved):
LGTM

the-technat (approved):
@mkilchhofer pinged me internally to have a look and I would confirm that fix is necessary since `maxUnavailable` and `minAvailable` are mutually exclusive.

tuunit:
> @mkilchhofer pinged me internally to have a look and I would confirm that fix is necessary since `maxUnavailable` and `minAvailable` are mutually exclusive.

Yes but null would be valid. It should work without kyverno and doesn't void the mutual exclusivity of those parameters as far as I know

tuunit:
Never the less this implementation is cleaner

PR #324, opened 2025-07-29
feat: Support spec.trafficDistribution for service

younsl:
## Background

spec.trafficDistribution field for Kubernetes Services provides fine-grained control over how traffic is distributed to service endpoints. trafficDistribution field became stable in Kubernetes 1.33, allowing administrators to specify traffic routing policies such as PreferClose, which prioritizes endpoints in the same zone or region as the client, reducing cross-zone network latency and data transfer costs.

## Changes

- Support [spec.trafficDistribution](https://kubernetes.io/docs/concepts/services-networking/service/#traffic-distribution) for service.


tuunit (approved):
LGTM

Issue #325, opened 2025-07-30
Feature Request: Support for Startup Probes


OliverLeighC:
I'm interested in adding support for [Kubernetes startup probes](https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#container-probes) to oauth2-proxy. My team uses Keycloak as our authentication provider, and we've noticed that if Keycloak is slow to start up, oauth2-proxy fails repeatedly on the OIDC Discovery step. This can cause unnecessary restarts and delays during deployments.

Having a configurable startup probe that can ping Keycloak (or another OIDC provider) before oauth2-proxy begins its normal operations would help ensure that dependencies are ready, improving reliability and startup behavior.

Proposed Solution:

- Add support for a startup probe in the oauth2-proxy deployment manifest.
- Allow the probe to check connectivity with the configured OIDC provider (e.g., Keycloak) before considering the container "started." (
- Make the probe optional and configurable, so users can enable it as needed.

Is this a feature you would be open to supporting? I made a PR with a simple implementation.

Thanks!

OliverLeighC:
Closing because as @tuunit  mentioned this makes more sense as an initContainer which is already supported with `Values.extraInitContainers`

PR #326, opened 2025-07-30
FR 325: Add support for startupProbe

OliverLeighC:
Related to https://github.com/oauth2-proxy/manifests/issues/325 

Would love to have this as an option. Tested via ` ct lint --config ct.yaml` as well as running `helm template` with different values and checking the rendered template.


helm/oauth2-proxy/values.yaml
```diff
@@ -281,6 +281,16 @@ readinessProbe:
   periodSeconds: 10
   successThreshold: 1
 
+startupProbe:
+  enabled: false
+  periodSeconds: 10
+  timeoutSeconds: 5
+  successThreshold: 1
+  failureThreshold: 3
+  httpGet: {}
```
tuunit, line 290: Shouldn't we decide what is the correct way to test that the startup is successful? Instead of letting the user decide on what to test for?

helm/oauth2-proxy/Chart.yaml
```diff
@@ -1,5 +1,5 @@
 name: oauth2-proxy
-version: 7.15.0
+version: 7.15.1
```
tuunit: As this is a new feature you need to increase the minor version not the patch

```suggestion
version: 7.16.0
```

tuunit: Furthermore, please add a changelog entry at the bottom of the Chart.yaml for artifacthub

tuunit (changes requested):

helm/oauth2-proxy/values.yaml
tuunit: Ah I just now read your issue on the topic. So you want to test outside dependencies like the IDP and/or the upstream 🤔 Isn't this normally a task for an init container?

tuunit: A startup probe should really be used for checking that the startup is successful and a container doesn't get flagged as RUNNING/READY before it's actual application can fully respond.

initContainers are usually used for the use case you are describing. We already have the waitForRedis init container and offer:

`Values.extraInitContainers`

OliverLeighC: I opted for the most flexible as a starting point, but that makes sense. What about the `.well-known/openid-configuration` endpoint on the `oidc_issuer_url` since that's whats used during discovery and is usually the first thing to fail if the auth provider isn't ready yet.

helm/oauth2-proxy/Chart.yaml
OliverLeighC: done!

helm/oauth2-proxy/values.yaml
OliverLeighC: > A startup probe should really be used for checking that the startup is successful and a container doesn't get flagged as RUNNING/READY before it's actual application can fully respond.
> 
> initContainers are usually used for the use case you are describing. We already have the waitForRedis init container and offer:
> 
> `Values.extraInitContainers`

An init container would make more sense, thanks for walking that through for me

OliverLeighC: This can be closed then, it's unlikely a startupProbe is needed since the startup itself is fairly quick.

PR #341, opened 2025-08-19
PDB fix with HPA enabled

### LINKED ISSUE #309: Pod disruption budget is not created when enabling HPA
`PodDisruptionBudget` is only generated [if `.Values.podDisruptionBudget.enabled` is true and `.Values.replicaCount` is greater than 1](https://github.com/oauth2-proxy/manifests/blob/4001c2385dc72990add71437b02e18d7ec7f8ffd/helm/oauth2-proxy/templates/poddisruptionbudget.yaml#L1).

[.Values.replicaCount is not set in the deployment manifest when using HPA](https://github.com/oauth2-proxy/manifests/blob/4001c2385dc72990add71437b02e18d7ec7f8ffd/helm/oauth2-proxy/templates/deployment.yaml#L14-L16), so the helm chart should take `.Values.autoscaling.minReplicas` in consideration as well when deciding if the `PodDisruptionBudget` should be rendered or not.

It's possible to workaround this by actually setting `replicaCount` to something that is greater than 1, but it's not very intuitive when using HPA.

pierluigilenoci:
Fixes #309

### DIFF
--- a/helm/oauth2-proxy/Chart.yaml
+++ b/helm/oauth2-proxy/Chart.yaml
@@ -1,5 +1,5 @@
 name: oauth2-proxy
-version: 8.1.0
+version: 8.1.1
 apiVersion: v2
 appVersion: 7.12.0
 home: https://oauth2-proxy.github.io/oauth2-proxy/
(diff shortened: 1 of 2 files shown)


pierluigilenoci:
@Fresa, considering this is a sensitive change, I'm waiting for a second opinion from another maintainer before merging.

tuunit (approved):

PR #345, opened 2025-08-21
Fix split brain in CI test

### LINKED ISSUE #344: Redis standalone values example split brain detection problem
This example https://github.com/oauth2-proxy/manifests/blob/main/helm/oauth2-proxy/ci/redis-standalone-values.yaml as mentioned in README.md file.

Have problem, which causes split brain check loop, high cpu usage and flood logs.

```
  splitBrainDetection:
    interval: 0

```

pierluigilenoci:
Fixes #344

### DIFF
--- a/helm/oauth2-proxy/Chart.yaml
+++ b/helm/oauth2-proxy/Chart.yaml
@@ -1,5 +1,5 @@
 name: oauth2-proxy
-version: 8.1.2
+version: 8.1.1
 apiVersion: v2
 appVersion: 7.12.0
 home: https://oauth2-proxy.github.io/oauth2-proxy/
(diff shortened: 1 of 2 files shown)


tuunit (approved):

helm/oauth2-proxy/Chart.yaml
```diff
@@ -1,5 +1,5 @@
 name: oauth2-proxy
-version: 8.1.1
+version: 8.1.2
```
tuunit: Do we really need a release for this?

pierluigilenoci: It was to please the pipeline.

PR #346, opened 2025-08-24
feat: Add support containers resize policy for deployment

younsl:
## Changes

- Add support spec.containers.resizePolicy field in deployment. This feature allows CPU and memory resources to be modified without pod restart or replacement.

## Reference

- [Resize CPU and Memory Resources assigned to Containers](https://kubernetes.io/docs/tasks/configure-pod-container/resize-container-resources/)


helm/oauth2-proxy/Chart.yaml
```diff
@@ -1,5 +1,5 @@
 name: oauth2-proxy
-version: 8.1.1
+version: 8.2.1
```
tuunit:
```suggestion
version: 8.2.0
```

You either upgrade major, minor or patch. Never two at once :)

https://semver.org/

helm/oauth2-proxy/templates/deployment.yaml
```diff
@@ -285,6 +285,10 @@ spec:
 {{- end }}
         resources:
 {{ toYaml .Values.resources | indent 10 }}
+{{- if .Values.resizePolicy }}
+        resizePolicy:
+{{ toYaml .Values.resizePolicy | indent 10 }}
+{{- end }}
```
tuunit:
```suggestion
{{- with .Values.resizePolicy }}
        resizePolicy: {{ toYaml . | nindent 10 }}
{{- end }}
```

I think this would be more elegant

helm/oauth2-proxy/Chart.yaml
tuunit: Furthermore, write a changelog entry for artifact hub at the bottom of this file

younsl: Sorry I missed. fixed it.

younsl:
@tuunit PTAL again.

Issue #355, opened 2025-09-27
Feature Request: Sign OCI Helm chart with keyless cosign


matheuscscp:
The provenance of the OCI Helm chart could be improved by signing it with cosign keyless:

https://docs.sigstore.dev/cosign/signing/overview/

Like this:

https://github.com/matheuscscp/mcp-oauth2-proxy/blob/aa8c2628517e77919c68d1dc1536e6b3d33dd8a2/.github/workflows/release.yml#L55-L64

PR #356, opened 2025-09-28
Sign chart with cosign

matheuscscp:
Closes: #355

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,3 +18,24 @@ Linting/validation uses the [helm/chart-testing tool](https://github.com/helm/ch
 ct lint --all --config ct.yaml
 ct install --all --config ct.yaml
 ```
+
+## Verify Signed Helm Charts
+
+With the introduction of cosign for signing artifacts you can verify the
+integrity of our artifacts using the following command:
+
+```
+VERSION=8.2.2
+cosign verify --certificate-oidc-issuer https://token.actions.githubusercontent.com \
+    --certificate-github-workflow-repository oauth2-proxy/manifests \
+    --certificate-github-workflow-name "Release Charts" \
+    --certificate-github-workflow-ref main \
+    --certificate-identity "https://github.com/oauth2-proxy/manifests/.github/workflows/release.yaml@main" \
+    "oci://ghcr.io/oauth2-proxy/manifests/charts/oauth2-proxy@${VERSION}" | jq
+```
+
+Note:
+
+We utilize cosign to sign and verify artifacts with the KEYLESS mode. To learn
+more about how keyless signing is done, visit the official documentation about 
+[Keyless Signatures](https://docs.sigstore.dev/cosign/signing/overview/#the-signing-witnessing-and-verifying-process).


.github/workflows/release.yaml
```diff
@@ -17,6 +17,9 @@ jobs:
         with:
           fetch-depth: 0
 
+      - name: Setup Cosign
+        uses: sigstore/cosign-installer@d7543c93d881b35a8faa02e8e3605f69b7a1ce62 # v3.10.0
```
tuunit: Please use the tag instead

.github/workflows/release.yaml
```diff
@@ -42,12 +45,14 @@ jobs:
...
         run: |
           shopt -s nullglob
           for pkg in .cr-release-packages/*; do
             if [ -z "${pkg:-}" ]; then
               break
             fi
-            helm push "${pkg}" oci://ghcr.io/${GITHUB_REPOSITORY_OWNER}/charts
+            repo="oci://ghcr.io/${GITHUB_REPOSITORY_OWNER}/charts"
+            helm push "${pkg}" "${repo}" |& grep Digest: | awk '{print $NF}' > digest.txt
+            cosign sign --yes "${repo}/oauth2-proxy@$(cat digest.txt)"
```
tuunit:
```suggestion
            cosign sign --yes "${repo}/${pkg}@$(cat digest.txt)"
```

shouldn't this be referencing the pkg variable?

tuunit: or does package contain the relative filesystem path? in that case we should do something like:

pkg=$(basename "${pkg%/}")

tuunit (changes requested):

.github/workflows/release.yaml
matheuscscp: The value of the variable `$pkg` is a string of the form `.cr-release-packages/oauth2-proxy-X.Y.Z.tgz`:

```bash
➜  oauth2-proxy git:(workspace) ✗ for pkg in ./tmp/*; do echo $pkg; done
./tmp/oauth2-proxy-1.0.0.tgz
➜  oauth2-proxy git:(workspace) ✗ 
```

`cosign` does not sign tags, it signs only digests. Helm concatenates the chart name (which is `oauth2-proxy` in this case) at the end of the `oci://` URL. `cosign` doesn't know that Helm do that, it expects the full URL. This `cosign` command is correct, trust me :hugs: 

See an example that actually executed here:

https://github.com/matheuscscp/mcp-oauth2-proxy/actions/runs/18079896658/workflow#L64

matheuscscp: Looks like `echo "${pkg##*/}" | sed 's/\.tgz$//' | sed 's/-[0-9].*$//'` can extract `oauth2-proxy` from `.cr-release-packages/oauth2-proxy-X.Y.Z.tgz`:

```bash
➜  oauth2-proxy git:(workspace) ✗ echo "${pkg##*/}" | sed 's/\.tgz$//' | sed 's/-[0-9].*$//'
oauth2-proxy
➜  oauth2-proxy git:(workspace) ✗ 
```

matheuscscp: Fixed, thanks!

.github/workflows/release.yaml
matheuscscp: Fixed, thanks!

.github/workflows/release.yaml
tuunit: `${pkg##*/}` already removes the file extension no need for the first sed

matheuscscp: Fixed again, thanks!

.github/workflows/release.yaml
```diff
@@ -42,12 +45,15 @@ jobs:
...
           shopt -s nullglob
           for pkg in .cr-release-packages/*; do
             if [ -z "${pkg:-}" ]; then
               break
             fi
-            helm push "${pkg}" oci://ghcr.io/${GITHUB_REPOSITORY_OWNER}/charts
+            chart_name=$(echo "${pkg##*/}" | sed 's/-[0-9].*$//')
+            repo="oci://ghcr.io/${GITHUB_REPOSITORY_OWNER}/charts"
+            helm push "${pkg}" "${repo}" |& grep Digest: | awk '{print $NF}' > digest.txt
+            cosign sign --yes "${repo}/${chart_name}@$(cat digest.txt)"
```
tuunit, line 59: Last but not least, don't we need to extend the permissions of the workflow to allow for `id-token: write` according to the documentation: https://github.com/sigstore/cosign-installer

matheuscscp: Most definitely, yes! Thanks for catching that! I also added comments explaining each permission, like I normally do for Helm OCI release workflows like this. Thanks!!

tuunit (approved):
LGTM now. I added some documentation to explain how to verify the signatures

PR #357, opened 2025-09-28
feat: add NetworkPolicy

matheuscscp:
This PR adds the optional creation of a `NetworkPolicy` resource with a fairly standard template used across open-source projects.


helm/oauth2-proxy/values.yaml
```diff
@@ -177,6 +177,14 @@ serviceAccount:
   automountServiceAccountToken: true
   annotations: {}
 
+# Network policy settings.
+networkPolicy:
+  create: false
+  ingress:
+    namespaces: []
+    #  - my-namespace-1
+    #  - my-namespace-2
```
tuunit: I don't think this is a good idea to enforce only a namespaced approach what about out of cluster connections?

tuunit: What about egress?

helm/oauth2-proxy/templates/networkpolicy.yaml
```diff
@@ -0,0 +1,35 @@
...
+    {{- with .Values.commonLabels }}
+    {{- toYaml . | nindent 4 }}
+    {{- end }}
+  {{- with .Values.commonAnnotations }}
+  annotations:
+    {{- toYaml . | nindent 4 }}
+  {{- end }}
+spec:
+  policyTypes:
+    - Ingress
```
tuunit, line 19:
```suggestion
  policyTypes:
    {{- if .Values.networkPolicy.ingress }}
    - Ingress
    {{- end }}
    {{- if .Values.networkPolicy.egress }}
    - Egress
    {{- end }}
```

helm/oauth2-proxy/templates/networkpolicy.yaml
```diff
@@ -0,0 +1,35 @@
...
+            matchLabels:
+              kubernetes.io/metadata.name: {{ .Release.Namespace }}
+        {{- range $ns := .Values.networkPolicy.ingress.namespaces }}
+        - namespaceSelector:
+            matchLabels:
+              kubernetes.io/metadata.name: {{ $ns }}
+        {{- end }}
+      ports:
+        - protocol: TCP
+          port: {{ .Values.httpScheme }}
```
tuunit: I would suggest we make this more generic and allow the users of the chart more control.
```suggestion
  {{- if .Values.networkPolicy.ingress }}
  ingress:
    {{- toYaml .Values.networkPolicy.ingress | nindent 4 }}
  {{- end }}
  {{- if .Values.networkPolicy.egress }}
  egress:
    {{- toYaml .Values.networkPolicy.egress | nindent 4 }}
  {{- end }}
```

helm/oauth2-proxy/values.yaml
tuunit:
```suggestion
  ingress: {}
  egress: {}
```

this way users can fully control which ingress and egress settings they want to use. this way they even if using the alpine version of our images they can guarantee egress only works for DNS and connecting to the upstream and IDP

matheuscscp: Feel free to propose a Helm values/template structure, I'm happy to implement it. This works for my use case :shrug:

matheuscscp:
> @matheuscscp can you have a look at my comments and bump the minor version of the Chart.yaml and update the changelog entry at the bottom of the same file

@tuunit All fixed, thanks!

tuunit (approved):

PR #359, opened 2025-09-30
fix: recommended setting for redis when using 1 replica

nleconte-csgroup:
Hello,

To use only 1 replicas, we need to update the redis configuration `min-replicas-to-write` to 0 otherwise we have the error message "NOREPLICAS Not enough good replicas to write".

Ref: https://stackoverflow.com/a/59737862

Note: It's well configured in the CI : https://github.com/oauth2-proxy/manifests/blob/3832a76bd98c42d787da8e2069ecdbdda45c4b5f/helm/oauth2-proxy/ci/redis-standalone-values.yaml#L29
It's just missing in the doc

### DIFF
--- a/helm/oauth2-proxy/values.yaml
+++ b/helm/oauth2-proxy/values.yaml
@@ -492,6 +492,7 @@ redis-ha:
   # Recommended:
   #
   # redisPassword: xxxxx
+  # When using replicas: 1, you MUST also set min-replicas-to-write: 0
   # replicas: 1
   # redis:
   #   config:


helm/oauth2-proxy/values.yaml
```diff
@@ -463,6 +463,9 @@ redis:
   #
```
tuunit, line 493: I think we should take the opportunity and change this section a bit more. Instead of having a single node cluster setup as the recommendation. We should reference the example files directory here. What do you think?

tuunit: @pierluigilenoci whats your stance on this?

pierluigilenoci: Anything that improves the readability or usefulness of the value file is welcome.

pierluigilenoci: @tuunit, do you think we should move forward with this fix?

PR #364, opened 2025-10-30
feat: adding HPA capabilities check

### LINKED ISSUE #363: HPAs on Kubernetes version 1.20.6 do not use the correct apiVersion
Kubernetes version 1.20.6 does not recognize the autoscaling/v2 api. I believe the autoscaling/v2 api was introduced in Kubernetes version 1.23.

A possible solution would be adding this to the _capabilities.tpl:

```
{{/*
Return the appropriate apiVersion for HorizontalPodAutoscaler object.
*/}}
{{- define "capabilities.horizontalPodAutoscaler.apiVersion" -}}
{{- if semverCompare "<1.23-0" ( .Values.kubeVersion | default .Capabilities.KubeVersion.Version ) -}}
{{- print "autoscaling/v2beta2" -}}
{{- else -}}
{{- print "autoscaling/v2" -}}
{{- end -}}
{{- end -}}
```
And updating the hpa.yaml with `apiVersion: {{ include "capabilities.horizontalPodAutoscaler.apiVersion" . }}`
Thank you!

Magnus010:
Checks if Kubernetes is older then 1.23-0 and sets the Horizontal Pod Autoscaler apiVersion to autoscaling/v2beta2 vs. the default of autoscaling/v2.

Tested in our extremely old Kubernetes cluster.

Fixes #363

### DIFF
--- a/helm/oauth2-proxy/Chart.yaml
+++ b/helm/oauth2-proxy/Chart.yaml
@@ -1,5 +1,5 @@
 name: oauth2-proxy
-version: 8.3.1
+version: 8.3.2
 apiVersion: v2
 appVersion: 7.12.0
 home: https://oauth2-proxy.github.io/oauth2-proxy/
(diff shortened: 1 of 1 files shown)


tuunit:
Hi @Magnus010,

please bump the patch version the Chart.yaml and update the changelog entry at the bottom of the same file

Magnus010:
Chart.yaml patch version bumped and Updated the changelog entry.

PR #367, opened 2025-11-12
feat!: separate image.repository into (image.registry / image.repository) and support .global.imageRegistry

FieteO:
## Description

- introduce `.global.imageRegistry` value to allow global configuration of the image registry
- split up `image.repository` into `image.registry` and `image.repository` to allow separate configuration
- use `image.registry` if `.global.imageRegistry` is not set

## Checklist:


helm/oauth2-proxy/values.yaml
```diff
@@ -75,7 +76,8 @@ alphaConfig:
   existingSecret: ~
 
 image:
-  repository: "quay.io/oauth2-proxy/oauth2-proxy"
+  registry: "quay.io"
+  repository: "oauth2-proxy/oauth2-proxy"
```
tuunit, line 79: this is a breaking change and would yield a bump to chart version `9.0.0`

pierluigilenoci: So also a migration guide must be added here https://github.com/oauth2-proxy/manifests/blob/main/helm/oauth2-proxy/README.md#upgrading-an-existing-release-to-a-new-major-version

helm/oauth2-proxy/Chart.yaml
```diff
@@ -1,5 +1,5 @@
 name: oauth2-proxy
-version: 8.3.3
+version: 8.4.0
```
pierluigilenoci:
```suggestion
version: 9.0.0
```

helm/oauth2-proxy/values.yaml
FieteO: Thanks for the review comments! I have now added a `9.0.0` section to the upgrade information in the README.
Please feel free to suggest different wording if you like

pierluigilenoci: Unfortunately, not all your commits are signed, which results in a DCO test failure.

FieteO: Ah, totally forgot about that. I will correct that tomorrow evening

FieteO: I have rebased the branch and signed off all commits now

tuunit: @FieteO  @pierluigilenoci To be honest, I still question this decision 🤔 
Is it really necessary to split the registry from the image?
If you use a different registry most of the time you don't have the same path for your images either and need to change it as well. So what is the benefit. I do get the desire to support global.registry but what is the use case here exactly?

helm/oauth2-proxy/README.md
```diff
@@ -100,6 +100,11 @@ Version 8.0.0 removes the dependency on the Bitnami Redis subchart and replaces
 
 Furthermore, you can read up on why this change was necessary in [Breaking changes in Bitnami Catalog #323](https://github.com/oauth2-proxy/manifests/issues/323)
 
+### To 9.0.0
+
+Version 9.0.0 introduces the `.global.imageRegistry` value that allows to centrally override the image registry that is used to pull images.
```
tuunit: For which purpose? Do you bundle OAuth2 Proxy together with other deployments? Like its done for promtheus kube stack and stuff, were half a dozen components are bundled into one chart?

Don't get me wrong. I do want support most use cases, I just want to have a better understanding of the user base for the helm chart and as you can imagine this change might be good for you but causes additional migration work for others that don't benefit from this change.

helm/oauth2-proxy/values.yaml
FieteO: In corporate environments you typically have some internal artifact repository like Sonatype Nexus or JFrog Artifactory that is mandatory to pull artifacts through.
You then have so called virtual or group repositories that aggregate all the allowed sources (`dockerhub`, `quay.io`,...) that can be pulled.
Since this is one central repository (i.e `artifactory.mycompany.com/my-project`) that all charts need to be configured for, it is really helpful to have `.global.imageRegistry` that you can configure once.
The alternative is that you have 30 different charts that all require their variant of overriding the image registry, which is something that you can avoid with this.

helm/oauth2-proxy/README.md
FieteO: Does my answer in the other thread address your concerns? Also keep in mind that this does not change the effective/templated values in the vanilla deployment.
People have to adjust their override for `.image.repository` (i.e from `artifactory.theircompany.com/their-virtual-repo/oauth2-proxy/oauth2-proxy` to a `image.registry` value of `artifactory.theircompany.com/their-virtual-repo`) if they use that, but that should be all from my perspective

tuunit:
@pierluigilenoci I reworded the breaking change notice in the README and Chart.yaml and the large changes in the README are due to the fact that the markdown table was unformatted and the global.imagePullSecret and global.imageRegistry was missing.

LGTM now

PR #399, opened 2026-03-24
feat: support templating in extraInitContainers

jwoloch:
## Description

Enable Helm template expressions in extraInitContainers by wrapping toYaml output with tpl function, allowing dynamic values from release metadata, config maps, and secrets.

## Checklist:


tuunit:
Could we stop opening a single PRs for each next element we want to add `tpl` to? Could we please just open one large PR for this?

pierluigilenoci:
Hi @jwoloch, thanks for this PR! Supporting templating in `extraInitContainers` is useful.

However, after testing locally, **this is a breaking change**. Wrapping `toYaml` with `tpl()` means that any existing `extraInitContainers` values containing literal `{{ }}` sequences will now be evaluated as Go template expressions and **fail with a parse error**.

**Reproduction:**
```yaml
extraInitContainers:
  - name: init-something
    image: busybox:1.36
    command: ["sh", "-c", "echo {{ some-annotation }}"]
```
- **Before (main):** renders `echo {{ some-annotation }}` literally ✅
- **After (this PR):** fails with `error calling tpl: cannot parse template [...]: bad character U+002D '-'` ❌

### Required changes

1. **Version bump:** The current bump to `10.4.3` conflicts with main (already at `10.4.3`). Since this is a breaking change, it needs a **major version bump** (`11.0.0`).

2. **Changelog kind** should be `changed` (not the current implicit "changed") and must explicitly state it's a breaking change:
   ```yaml
   - kind: changed
     description: |
       BREAKING: extraInitContainers values are now evaluated as Go templates
       via tpl(). Values containing literal {{ }} must be escaped using
       {{ "{{" }} and {{ "}}" }}.
   ```

3. **Use `$` instead of `.`** for the template context — for consistency with the rest of the chart:
   ```diff
   -        {{- tpl (toYaml .Values.extraInitContainers) . | nindent 6 }}
   +        {{- tpl (toYaml .Values.extraInitContainers) $ | nindent 6 }}
   ```
   While `.` works here (not re-scoped at this point), `$` is the pattern used everywhere else in this chart (`deploymentAnnotations`, `ingress.annotations`, `cookieName`, etc.).

4. **Add a CI test value file** to validate templating works and escaping is documented:
   ```yaml
   # ci/extra-init-containers-tpl-values.yaml
   extraInitContainers:
     - name: init-config
       image: "busybox:{{ $.Chart.AppVersion }}"
       command: ["sh", "-c", "echo $(RELEASE_NAME)"]
       env:
         - name: RELEASE_NAME
           value: "{{ $.Release.Name }}"
   ```

5. **Add migration guidance** in the PR description for users who have literal `{{ }}` in their `extraInitContainers` values.

Let me know if you have questions or need help with these changes!

pierluigilenoci:
Following up on my previous comment — I did further validation against a live Kubernetes cluster and want to clarify:

Unlike PR #400 (where all affected fields have strict character validation that already prevents `{{ }}`), **this PR does affect fields that accept arbitrary strings**. Specifically:

- `command[]` — accepts any string ✅ (verified: `kubectl apply --dry-run=client` passes with `{{ }}` in command)
- `args[]` — accepts any string ✅
- `env[].value` — accepts any string ✅

So a user could legitimately have something like:
```yaml
extraInitContainers:
  - name: init-config
    image: busybox:1.36
    command: ["sh", "-c", "envsubst '{{ .MY_VAR }}' < /tpl/config > /config/app.yaml"]
```

This works today on main and would **break** after this PR lands.

So my original assessment stands — this is a real breaking change for `extraInitContainers` specifically. The required changes from my previous comment still apply:

1. **Major version bump** (currently conflicts with main anyway)
2. **Use `$` instead of `.`** for consistency
3. **Add a CI test value file** exercising the templating
4. **Document the breaking change** in changelog and PR description with escaping guidance (`{{ "{{" }}`)

Let me know if you have questions!

PR #402, opened 2026-03-29
fix: configmap generation with `no-config` mode for external secret management

tuunit:
## Description

Add back support for running without generated/mounted ConfigMap when `alphaConfig.enabled=false` and `forceLegacyConfig=false`. This enables users to manage oauth2-proxy configuration entirely via external secrets or other means as was previously possible as mentioned by https://github.com/oauth2-proxy/manifests/pull/385#discussion_r3001082837

- Add `no-config` legacy-config mode
- Skip ConfigMap generation and volume mounts in `no-config` mode
- Add `mode` label to generated ConfigMaps for debugging
- Update values.yaml documentation

## Checklist:

- [X] I have bumped the version in the Chart.yaml according to [Semantic Versioning](https://semver.org).
- [X] I have updated the documentation/CHANGELOG at the bottom of the Chart.yaml
- [X] I have [signed off](https://github.com/apps/dco) all my commits.


pierluigilenoci:
Hi! I've opened #404 as a minimal, surgical fix for the same issue reported by @MattiasGees.

The key difference is that #404 only adds a `no-config` branch to the existing `legacy-config.mode` helper (2 lines in `_helpers.tpl`) and gates the ConfigMap generation, `--config` flag, volume mount, and volume definition on it — with zero renames, no new helpers, and no behavioral changes for existing configurations.

I think the structural improvements in this PR (alpha-config helpers, deprecation guards, better naming) are valuable, but they'd be better suited as a separate follow-up refactor PR after the core fix lands, to keep the blast radius small and review straightforward.

pierluigilenoci:
Following up on my previous comment — I've now split the work across two PRs:

- **#404** — Minimal bug fix: adds `no-config` legacy mode so users managing config externally (CSI SecretStore Driver, etc.) are no longer broken by #385. Surgical change, no renames, no breaking changes.

- **#405** — Structural improvements cherry-picked from this PR: `alpha-config.source` and `alpha-config.name` helpers, deprecation guards for invalid `existingConfig`/`existingSecret` combinations, simplified `secret-alpha.yaml` and `deployment.yaml` volume logic, CI test for `alphaConfig.existingSecret`.

Intentionally left out from #405:
- Renaming `legacy-config.mode` → `legacy-config.source` (breaking for custom templates)
- Always-on debug annotations (`legacy/source`, `alpha/source`)
- `legacy-config.enabled` helper (thin wrapper, not needed)

This way the bug fix can land quickly (#404) and the refactor improvements can be reviewed independently (#405). Thanks for the original work on this PR — it was a great starting point!

tuunit:
@pierluigilenoci I have to disagree with all three points that have been left out in the split. I don't understand why the split into two more PRs is necessary. We could just work in this PR to figure out a final solution.

I think renaming to source will not breaking anything. The AI isn't aware of when the release was done and in my opinion no one will have built custom templating based on the new helpers in the last 3 days...

Furthermore, why wouldn't we want the annotations for the type of config used? It a quick insight and doesn't hurt anyone.


One other possible route we could take, is revert the original PR and take more time to fully rework this mess to have a clean state even if it means introducing a breaking change

PR #417, opened 2026-06-23
feat(GatewayAPI): Add timeouts for httpRoute

Momotoculteur:
## Description

Add support for configuring per-rule HTTPRoute timeouts in the Gateway API HTTPRoute template

Each entry under gatewayApi.rules now accepts an optional timeouts block exposing the two Gateway API fields:

- timeouts.request — maximum duration for the gateway to respond to a client request
- timeouts.backendRequest — timeout for an individual request from the gateway to the backend

Both values are Gateway API Duration strings (e.g. "30s", "500ms")

The field is rendered only when set, so existing installations are unaffected and the default rule behaviour is unchanged

## Changes:

- templates/httproute.yaml: render timeouts for each rule when defined
- values.yaml: document the new timeouts field in the gatewayApi.rules example
- ci/gateway-api-timeouts.yaml: new CI values file covering a rule with custom request/backendRequest timeouts

## Checklist:

- [x] (Optional) I have implemented helm tests for new feature flags

### DIFF
--- a/helm/oauth2-proxy/Chart.yaml
+++ b/helm/oauth2-proxy/Chart.yaml
@@ -1,5 +1,5 @@
 name: oauth2-proxy
-version: 10.7.0
+version: 10.8.0
 apiVersion: v2
 appVersion: 7.15.3
 home: https://oauth2-proxy.github.io/oauth2-proxy/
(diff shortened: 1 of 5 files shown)


pierluigilenoci:
Hi @Momotoculteur, thanks for this PR! Adding per-rule timeouts to HTTPRoute is a useful addition — the Gateway API spec supports it and it fills a real gap.

CI approved and running. A few small things to address:

### 1. Minor typo in README

Line with the example has an unmatched closing parenthesis:

```
e.g. `"10s"`, `"500ms"`). See the...
                   ^
```

Should be:

```
e.g. `"10s"`, `"500ms"`). See the...  ← remove the stray )
```

### 2. Version bump conflict

You bumped to `10.8.0`, which conflicts with two in-flight PRs (#414, #415 — both at 10.8.0). We'll reconcile at merge time, no action needed from you now.

### 3. CI test values indentation

In `ci/gateway-api-timeouts.yaml`, the `matches` list item is not indented under the rule:

```yaml
  rules:
  - matches:         ← should be indented under the rule, not at same level
    - path:
```

This should be:

```yaml
  rules:
    - matches:
        - path:
            type: PathPrefix
            value: /oauth2
      timeouts:
        request: 30s
        backendRequest: 10s
```

Otherwise the implementation looks correct — `toYaml .timeouts | nindent 6` is the right pattern and is consistent with how `filters` and `matches` are rendered in the same template. Let me know if you have questions!

